PR #240, opened 2026-04-06
fix(mobile): author-only Edit visibility after auth hydration

mustafaocakxyz:
Add isRecipeAuthor helper (session id vs recipe.author.id with numeric coercion). Recipe detail shows Edit only when isReady, authenticated, and author matches. Reuse helper in RecipeEdit forbidden check.


CaganIslam (approved):
LGTM

PR #245, opened 2026-04-06
feat(mobile): improved search empty states

mustafaocakxyz:
Add reusable EmptyState component and use it in Search for both pristine state (no query/region yet) and no-results state with clear actions.


CaganIslam:
Looks good to me

PR #262, opened 2026-04-07
style(mobile): align UI theme with web

mustafaocakxyz:
Aligned UI theme (fonts, colors and etc.) with web.


ErenCanOzkaya (changes requested):
Home page looks good, please apply the style changes to all pages, then I believe we can merge it into the main

PR #401, opened 2026-04-28
feat: add comment and question api for recipes

roboticrustacean:
Created api for comments and questions for POST and DELETE.


laternis (dismissed):
Comment model and the basic endpoints look solid overall. A few things I'd want sorted before merging though:
                                                                                                                                                                
  Tests are missing. The recipes app already has four test modules (tests.py, tests_custom_submission_api.py, tests_lookup_apis.py, tests_permissions.py), so a
  tests_comments.py would fit right in. At minimum I think we need: anon GET/POST/DELETE behaviour, non-author getting 403 on delete, author getting 204, the   
  reply flow via parent_comment, COMMENT vs QUESTION typing, and pagination.
                                                                                                                                                                
  Recipe.qa_enabled is being ignored. The Recipe model has a qa_enabled flag but the comments action in views.py never checks it, so a user can still post      
  type=QUESTION even when the recipe owner has Q&A turned off. Something like this on the POST branch should cover it:
                                                                                                                                                                
  if request.data.get('type') == 'QUESTION' and not recipe.qa_enabled:
      return Response({'detail': 'Q&A disabled for this recipe.'}, status=403)                                                                                  
                                                                                                                                                                
  parent_comment isn't validated. It's writable on CommentSerializer but nothing stops a user from passing a comment id that belongs to a different recipe,     
  which leaves replies in an inconsistent state. A simple check in validate would fix it:                                                                       
                                                                                                                                                                
  def validate(self, attrs):                   
      parent = attrs.get('parent_comment')     
      recipe = self.context.get('recipe')
      if parent and recipe and parent.recipe_id != recipe.id:                                                                                                   
          raise serializers.ValidationError({'parent_comment': 'Must belong to the same recipe.'})
      return attrs                                                                                                                                              
                   
  You'll also need to pass context={'recipe': recipe, 'request': request} when instantiating the serializer in the view.                                        
                   
  Once these three are addressed I'm happy to approve.

roboticrustacean:
@laternis in commit 5974121 i addressed your review points. can you check if this is safe to merge now?

laternis:
LGTM — all three concerns addressed. The qa_enabled gate, the cross-recipe parent_comment validation, and the test coverage are all in

laternis (approved):
LGTM — all three concerns addressed. The qa_enabled gate, the cross-recipe parent_comment validation, and the test coverage are all in

Issue #335, opened 2026-04-18
M4-04 — Creator notification on new question


ErenCanOzkaya:
## Description
Emit a push notification and in-app notification to the recipe author whenever a new question is posted on their recipe. Backend publishes the event; web and mobile consume the notification stream and render it.

## Motivation / Context
Authors need to know when someone asks about their recipe — otherwise Q&A stagnates.

## Deliverables
- Backend event emission on new question
- Web + mobile in-app notification render
- Push notification delivery

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.4.2
- Depends on #M4-01

roboticrustacean:
PR #402 handles this issues backend side once merged.

laternis:
done

PR #402, opened 2026-04-28
feat: implement creator notification system for new questions #335

roboticrustacean:
Create notification system for questions on top of the comment/question system implemented in pr #401


laternis:
LGTM — signal-level coverage is solid and the self-notification guard plus the graceful fallback when the Expo SDK is missing both look right.

Issue #336, opened 2026-04-18
M4-05 — Reply notification to asker


ErenCanOzkaya:
## Description
Push + in-app notification to the original asker when their question receives a reply.

## Motivation / Context
Closes the Q&A loop — askers are notified of answers.

## Deliverables
- Reply event wired through notification pipeline
- Client-side notification rendering

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.4.3
- Depends on #M4-01, #M4-04

roboticrustacean:
Created the backend and tested manually in postman as well as unit tested in PR #403

PR #403, opened 2026-04-29
feat: implement reply notification to asker #336

roboticrustacean:
Created backend side of reply notification system for issue #336


laternis:
LGTM — clean mirror of the question-notification flow, and the four tests cover the cases that matter (self-reply, reply-to-non-question, no duplicate on     
  update).

Issue #337, opened 2026-04-18
M4-06 — Helpful / upvote backend


ErenCanOzkaya:
## Description
Vote model (userId, commentId, unique constraint) + aggregation endpoint returning total helpful count per comment. One vote per user per reply; idempotent toggle.

## Motivation / Context
Lets users surface the most useful answers.

## Deliverables
- Vote model + migration
- Toggle + aggregation endpoints

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.4.4
- Depends on #M4-01

roboticrustacean:
#404 closes this when merged.

laternis:
done

PR #404, opened 2026-04-29
feat: implement helpful / upvote backend #337

roboticrustacean:
Created vote system backend as per #337


laternis (changes requested):
The annotate-based counting and the toggle pattern via get_or_create/delete look good — way better than per-row counts. A few things before this one:
                                                                                                                                                                
  Missing tests. #402 and #403 both shipped with their own test classes, this one drops back to zero. For a write endpoint that mutates state and a serializer  
  that now returns helpful_count/has_voted, we should at least cover: anon POST → 401, first POST creates (201), second POST from same user toggles off (200),  
  separate users count independently, helpful_count reflects the total, and has_voted returns true only for the caller.                                         
                                                                                                                                                                
  Self-voting isn't blocked. Right now an author can upvote their own comment/answer. Depending on what we agreed for #337, that's usually disallowed — worth   
  deciding explicitly and short-circuiting in the action if so:
                                                                                                                                                                
  if comment.author_id == request.user.id:
      return Response({'detail': "You can't vote on your own comment."}, status=400)

roboticrustacean:
9dcd3d5038a5d0f0a7ef59b8d60f595f2af8bcc0 addresses your review points. note that I still allowed self vote as that is not decided elsewhere in our project, yet it's common for similar platforms like reddit to allow self voting. can you re review @laternis

Issue #339, opened 2026-04-18
M4-08 — Messaging data model & API


ErenCanOzkaya:
## Description
1–1 thread model: Thread, Message, ThreadParticipant, read state per participant, paging cursor, last-message index for the inbox list. Endpoints for create/send/read/list.

## Motivation / Context
Foundation for M4 private messaging (M4-09, M4-10, M4-11).

## Deliverables
- Thread/Message/ThreadParticipant models
- Create / send / read / list endpoints
- Inbox ordering index

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.4.5

roboticrustacean:
#406 will close this once merged.

laternis:
done

PR #406, opened 2026-04-29
feat: implement messaging API and schema

roboticrustacean:
Create messaging backend schema and APIs and test manually in postman and unit tests as per #339


laternis (changes requested):
This is a chunky one but the schema + soft-delete + idempotent thread creation all hang together well. Tests are scoped to the right flows, idempotency check
  returning 200 vs 201 is a nice touch, and auto-advancing the sender's read cursor on send is a small but considerate detail.                                  
                                                                                                                                                                
  A few real concerns before this lands:                                                                                                                        
                                                                                                                                                                
  Inbox is N+1 on unread_count. ThreadSerializer.get_unread_count runs two queries per thread (one for the participant lookup, one to count messages). On a     
  50-thread inbox that's ~100 queries before we even count _other_participant's lookups. The fix is to annotate the queryset once:                              
                                                                                                                                                                
  me_participant = ThreadParticipant.objects.filter(thread=OuterRef('pk'), user=request.user)
  qs = qs.annotate(                                                                                                                                             
      my_last_read=Subquery(me_participant.values('last_read_at')[:1]),
      unread_count=Count('messages', filter=Q(messages__is_deleted=False) & Q(messages__created_at__gt=F('my_last_read'))),                                     
  )                                                                                                                                                             
  …or similar. Worth doing now; it'll be much harder to retrofit once the frontend depends on the response shape.                                               
                                                                                                                                                                
  _other_participant ignores the prefetch. get_queryset prefetches participants__user, but _other_participant calls obj.participants.exclude(...).first() which
  fires a fresh query. Iterating in Python preserves the prefetch:                                                                                              
                   
  def _other_participant(self, obj):                                                                                                                            
      me = self.context['request'].user                                                                                                                         
      return next((p for p in obj.participants.all() if p.user_id != me.id), None)
                                                                                                                                                                
  Idempotent thread create has a TOCTOU race. Two concurrent POSTs from user A with the same other_user_id can both fall through the "existing" lookup and      
  create two thread rows. Wrapping the create branch in transaction.atomic() with a select_for_update on a deterministic key would close it. Not catastrophic at
   our scale but cheap to fix.                                                                                                                                  
                                               
  Body validation is missing. MessageWriteSerializer accepts empty / whitespace-only bodies straight through. A validate_body that strips and rejects empty     
  would prevent garbage messages.
                                                                                                                                                                
  seed_messaging has buggy / dead code. The first get_or_create(participants__user=cook, ...) is referencing a reverse relation in the lookup which isn't really
   meaningful, and the returned thread variable is never used because the code immediately re-checks via a separate filter and creates new objects. Can drop the
   first block entirely:                                                                                                                                        
                   
  def seed_messaging(self):                    
      cook = User.objects.get(username='cook')
      student = User.objects.get(username='student')                                                                                                            
      if Thread.objects.filter(participants__user=cook).filter(participants__user=student).exists():
          return                                                                                                                                                
      thread = Thread.objects.create()
      ...

PR #416, opened 2026-04-30
fix(mobile): handle paginated recipes list response

### LINKED ISSUE #413: Mobile: home recipes list is empty after backend pagination change
## Description
The Recipes section on the mobile Home screen is empty even though the backend has 17 recipes seeded. Affects Home and any other screen that calls `fetchRecipesList` (StoryCreate, StoryEdit, UserProfile).

## Root cause
Backend `RecipeViewSet` was switched to `StandardResultsSetPagination` as part of the comment/question API merge. `GET /api/recipes/` now returns:

```json
{ "count": 17, "next": null, "previous": null, "results": [ ... ] }
```

…but `app/mobile/src/services/recipeService.ts::fetchRecipesList` does `apiGetJson<any[]>` and `Array.isArray(data) ? data : []`, so the flat-array fallback returns `[]` and the UI shows nothing.

Other mobile callers like `fetchRecipeById` are unaffected because the detail endpoint still returns a single object.

## Expected
Home shows the seeded recipes under "More recipes". Story create/edit recipe pickers and UserProfile recipe list also populate.

## Reproduce
1. Lokal backend (post comment-API merge) — `python manage.py runserver`.
2. Mobile app on emulator with `EXPO_PUBLIC_API_URL=http://10.0.2.2:8000`.
3. Open Home → no recipes.
4. `curl http://localhost:8000/api/recipes/` returns paginated `{ count, results }` shape.

## Fix
`fetchRecipesList` should accept either a flat array or `{ results: [] }`. Patch shipped on branch `feat/mobile/comment-ui-334` (will be picked up by that PR):

```ts
const data = await apiGetJson<any>(`/api/recipes/`);
const list: any[] = Array.isArray(data)
  ? data
  : data && Array.isArray(data.results)
    ? data.results
    : [];
```

## Notes
- Verify pagination is intentional on the recipes list (web frontend may also be affected — needs a separate check).
- Long-term we may want a typed `Paginated<T>` helper in `httpClient.ts` since `/api/recipes/:id/comments/` already paginates the same way.

CaganIslam:
## Summary
`fetchRecipesList` was returning empty after the recipes endpoint switched to a paginated `{ count, next, previous, results }` payload (introduced with the comment-API merge). It used `apiGetJson<any[]>` and `Array.isArray(data) ? data : []`, so the new shape collapsed to `[]` and broke the home recipes list, story-link picker, and profile recipe list.

Now accepts both flat array and paginated payload — falls back to `data.results` when present.

Closes #413

## Test plan
- [ ] Home shows recipes again under "More recipes"
- [ ] Story create/edit recipe picker populates
- [ ] UserProfile recipe tab populates
- [ ] No regression when backend returns a flat array

### DIFF
--- a/app/mobile/src/services/recipeService.ts
+++ b/app/mobile/src/services/recipeService.ts
@@ -110,8 +110,13 @@ export async function fetchRecipesList(): Promise<
     image?: string | null;
   }[]
 > {
-  const data = await apiGetJson<any[]>(`/api/recipes/`);
-  return (Array.isArray(data) ? data : []).map((r) => {
+  const data = await apiGetJson<any>(`/api/recipes/`);
+  const list: any[] = Array.isArray(data)
+    ? data
+    : data && Array.isArray(data.results)
+      ? data.results
+      : [];
+  return list.map((r) => {
     const reg = r.region;
     const regionLabel =
       reg == null


mustafaocakxyz (approved):
Nice fix, good job.

PR #430, opened 2026-05-02
feat: implement backend support for contactability toggle and enable Cors

### LINKED ISSUE #342: M4-11 — Contactability toggle
## Description
User setting that controls whether others may open new threads. Backend enforces the preference server-side regardless of UI state.

## Motivation / Context
User privacy / opt-out from unsolicited DMs.

## Deliverables
- Preference field on user model
- Settings UI on web + mobile
- Server-side enforcement on thread creation

## Acceptance Criteria

## References
- Requirement 3.4.6
- Depends on #M4-08

roboticrustacean:
Closes #342 . This implements the backend side of contactability toggle.


laternis (commented):
Reviewed the diff. Feature core is solid — toggle is gated server-side at the right entry point, default `True` preserves existing UX, and the existing-thread-not-frozen behavior is the right semantic and is tested. Nice work.

Two things I'd like to see fixed before merging:

### 1. `CORS_ALLOW_ALL_ORIGINS = True` will land on production
In `config/settings.py`, the comment says "Only for development" but the setting applies in every environment. Once this hits the Vultr deploy, any origin can read API responses. Suggest gating on `DEBUG` or using an explicit allowlist:

```python
if DEBUG:
    CORS_ALLOW_ALL_ORIGINS = True
else:
    CORS_ALLOWED_ORIGINS = os.getenv('CORS_ALLOWED_ORIGINS', '').split(',')
```

Also: the CORS work is unrelated to #342 (contactability toggle). Ideally it would land as its own PR so each piece can be reverted independently if anything regresses.

Minor on CORS: `corsheaders.middleware.CorsMiddleware` should sit right after `SecurityMiddleware` per the package's docs — current placement still works for typical cases but isn't the recommended order.

### 2. `is_contactable` is being threaded through `CulturalProfileUpdateSerializer`
That serializer is named for cultural-onboarding fields. There's an existing test (`test_patch_ignores_non_cultural_fields`) that explicitly guards "PATCH only touches cultural fields" — adding `is_contactable` to it silently breaks that contract.

Two options, either is fine:
- Rename to `UserPreferencesUpdateSerializer` (or similar) and update the stale test name/docstring, OR
- Keep the cultural serializer focused and add a sibling `ContactabilityUpdateSerializer`.

It would also be nice to add a regression test confirming a PATCH like `{'role': 'admin', 'is_contactable': false}` still rejects the role escalation, so we know the new field didn't open a privilege-escalation gap via the serializer.

### Minor (non-blocking)
- The 403 test asserts `'disabled' in res.data['detail']` — substring matches couple tests to copy. Prefer status-only or full-string equality.
- Consider a note in the field's `help_text` about the asymmetry: a contact-disabled user can still *initiate* threads outbound; only inbound new-thread creation is blocked. Likely the intended behavior, just worth documenting.

Once 1 and 2 are addressed I'm happy to approve. Thanks!

roboticrustacean:
I've addressed the feedback from the previous review in commit 0c0ef9c64224d493c57befee08b6fa9819c6f267:

CORS: Gated CORS_ALLOW_ALL_ORIGINS on DEBUG, implemented CORS_ALLOWED_ORIGINS for production, and corrected the middleware order.
Serializer: Renamed CulturalProfileUpdateSerializer to UserPreferencesUpdateSerializer and updated the tests.
Security: Added a regression test to confirm sensitive fields (role, is_staff) are ignored during PATCH.
Minor: Updated the 403 assertion to a full string match and clarified the is_contactable help_text.
Ready for another look @laternis!

Emirhan777 (commented):
Correctly adds is_contactable boolean field. The core logic is correct, migration is backward compatible, access control is properly enforced, and there are good tests.

Emirhan777 (approved):
Correctly adds is_contactable boolean field. The core logic is correct, migration is backward compatible, access control is properly enforced, and there are good tests.

Issue #348, opened 2026-04-18
M4-17 — Daily cultural content backend


ErenCanOzkaya:
## Description
Service that assembles daily facts, quizzes, and calendar/holiday-themed items, personalized using the onboarding profile. Content team provides seed items; backend schedules and serves.

## Motivation / Context
Daily re-engagement hook.

## Deliverables
- Content model for daily items
- Scheduler + personalization layer
- Seed content batch from Content team

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Draft requirement D.7
- Depends on #M4-12

PR #443, opened 2026-05-03
feat: implement preference-driven personalization ranking logic and API integration

### LINKED ISSUE #345: M4-14 — Preference-driven personalization
## Description
Feed, Explore, map discovery, and recommendations read the onboarding profile and rank content accordingly. Backend exposes ranking input; clients consume ranked output.

## Motivation / Context
Turns the profile data from M4-12/13 into actual personalized surfaces.

## Deliverables
- Ranking component on backend
- Feed / Explore / Map / Recs consume ranked output

## Acceptance Criteria

## References
- Draft requirement D.5.2
- Depends on #M4-12

roboticrustacean:
Closes #345 : Implements the backend ranking component for personalization based on user onboarding profiles.

Added scoring system in apps/common/personalization.py that matches content with user interests, regional ties, and dietary/event preferences.
Applied ranking to Feed, Explore, Map discovery, and Global Search across Recipes, Stories, and Cultural Content.
Includes comprehensive unit and integration tests (29 passing cases) to ensure correct content ordering and metadata exposure (rank_score, rank_reason).

### DIFF
--- a/app/backend/apps/common/pagination.py
+++ b/app/backend/apps/common/pagination.py
@@ -1,0 +1,6 @@
+from rest_framework.pagination import PageNumberPagination
+
+class StandardResultsSetPagination(PageNumberPagination):
+    page_size = 10
+    page_size_query_param = 'page_size'
+    max_page_size = 100
--- a/app/backend/apps/common/personalization.py
+++ b/app/backend/apps/common/personalization.py
@@ -15,6 +15,11 @@ RANK_REASONS = {
     'cultural': 'cultural_match',
 }
 
+# Magic Weights Rationale:
+# Regional matches (40) are prioritized as the strongest signal of direct relevance.
+# Dietary preferences (30) follow, ensuring content aligns with essential lifestyle choices.
+# Event interests (25) capture specific occasion-based relevance.
+# Cultural text matches (15) provide a broader but noisier signal for discovery.
 WEIGHTS = {
     'regional': 40,
     'dietary': 30,
@@ -53,19 +58,25 @@ def attach_rank(obj, score, reason):
     return obj
 
 
-def rank_items(items, user, scorer, *, recency_attr='created_at'):
+def rank_items(items, user, scorer, *, recency_attr='created_at', personalize=True, scorer_kwargs=None):
+    """Rank items based on user profile.
+    
+    If personalize=False, or user has no profile terms, returns items as-is
+    to allow database-level pagination/lazy evaluation.
+    """
+    if not personalize or not has_profile_terms(user):
+        return items
+
+    if scorer_kwargs is None:
+        scorer_kwargs = {}
+
+    # Materialize items for ranking (usually with a soft cap from the view)
+    items = list(items)
     ranked = []
-    use_profile = has_profile_terms(user)
     for index, item in enumerate(items):
-        if use_profile:
-            score, reason = scorer(item, user)
-        else:
-            score, reason = 0, None
+        score, reason = scorer(item, user, **scorer_kwargs)
         attach_rank(item, score, reason)
         ranked.append((item, index))
-
-    if not use_profile:
-        return [item for item, _ in ranked]
 
     ranked.sort(
         key=lambda pair: (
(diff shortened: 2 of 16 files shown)


laternis (commented):
Reviewed the diff. The shared `apps/common/personalization.py` module is well-factored, the anonymous + empty-profile fallback is explicit and correct, and replacing the inline `_score` from #348 with the shared `score_cultural_content` is a clean win. Tie-breaking is deterministic. Tests are real, not mock-heavy. Nice work overall.

A few things I'd like to see fixed before merging:

### Critical: 4 list endpoints hydrate the full table before pagination

`apps/recipes/views.py` `list` does:
```python
items = rank_items(list(queryset), request.user, score_recipe)
page = self.paginate_queryset(items)
```

`list(queryset)` materializes every published recipe into Python, scores all of them, then paginates. PostgreSQL can't use indexes for sorting after that. With even a few thousand recipes this becomes a P95 cliff. Same pattern in `StoryViewSet.list`, `RecommendationsView`, and `GlobalSearchView`.

Minimum fix: for anonymous / empty-profile users (where `rank_items` is a no-op anyway), call `super().list()` so the queryset stays lazy and DB-paginated:

```python
def list(self, request, *args, **kwargs):
    if not has_profile_terms(request.user):
        return super().list(request, *args, **kwargs)
    queryset = self.filter_queryset(self.get_queryset())
    items = rank_items(list(queryset), request.user, score_recipe)
    ...
```

Better long-term: push scoring into SQL annotations or apply a soft cap (`items[:500]`) before scoring. The `RecommendationsView` is the worst case — it hydrates ALL recipes + ALL stories regardless of `?limit=`.

### Important: `/api/stories/` response shape changes for anonymous users

`apps/stories/views.py` overrides `list` and returns `Response(serializer.data)` (a bare list) instead of the paginated `{count, results, ...}` envelope. If the frontend reads `response.data.results` from `/api/stories/` anywhere, this PR breaks it. Worth confirming with the frontend team before merging.

### Important: dead `?surface=` param

`GlobalSearchView` accepts `?surface=feed|explore|map|recs` but the value is echoed back unchanged and doesn't affect ranking or filtering. Either branch on it (e.g., `surface=map` should join region data; `feed` should weight recency higher) or drop it — currently it's misleading API surface.

### Important: no opt-out

There's no way to disable ranking on `/api/recipes/` or `/api/stories/` for an authenticated user. Suggest `?personalize=0` as an escape hatch — useful for the frontend's "browse all" mode and for backend QA.

### Minor / observations
- The English-only stop-word list in `_significant_tokens` (`'cuisine', 'food', 'dish', ...`) means a user with `cultural_interests=['street food']` will match content that only contains `'street'`. False positive risk for short common words. Consider only stripping the stop-word *after* a full-phrase match has failed, or document the behavior.
- `apps/search/views.py` re-imports `apply_recipe_filters` from `apps.recipes.views` — cross-app reach into another app's view module. Lift `apply_recipe_filters` into `apps/recipes/filters.py` to break that coupling.
- Magic weights `40/30/25/15` would benefit from a one-line rationale in the docstring (regional > dietary > event > cultural-text).
- Test gaps: anonymous user explicitly hitting list endpoints and asserting `rank_score=0`, no-match user fallback, pagination + ranking integration. Also worth running `python manage.py test apps.cultural_content` against this branch since `_score` was replaced (the math changed even though the ordering may still align with existing test fixtures).

Big picture: the personalization logic itself is good. Once the pagination cliff is addressed and the story response shape is preserved (or confirmed safe with the frontend), I'm happy to approve. Thanks!

roboticrustacean:
I have addressed the PR feedback with the following improvements in 9cacce931db909ba0533b2a7e1a037bd4fe7050a:

Performance: Resolved the pagination cliff by skipping ranking for unauthenticated users and applying a soft cap (500 items) for ranked results.
Consistency: Standardized the response shape for /api/stories/ with pagination.
Decoupling: Extracted filter logic into apps.recipes.filters.py to eliminate cross-app coupling.
Logic Refinements:
Documented magic weights rationale.
Improved stop-word handling to prioritize full-phrase matches.
Implemented ?personalize=0 opt-out and ?surface= aware weighting.
Tests: Added a new regression test suite (apps.common.tests.test_personalization_refinement) and updated existing tests to reflect pagination changes. All backend tests are now passing. 
Can you re-review @laternis ?

Emirhan777:
LGTM.

PR #457, opened 2026-05-08
feat: implement map discovery backend infrastructure and endpoints #381

### LINKED ISSUE #381: #M5-15 — Map discovery backend
## Description
Region-tagged content index + endpoints to query content bounded by geographic/cultural region.

## Motivation / Context
Foundation for the map discovery screens (#M5-16, #M5-17).

## Deliverables
- Region tag model + index
- Region-bounded query endpoints

## Acceptance Criteria

## References
- Draft requirement D.2.1

roboticrustacean:
Closes #381 
This PR implements the backend foundation for map-based content discovery. It introduces geographic metadata to regions and provides a set of viewport-aware endpoints for future map integrations (e.g., Resium/Mapbox).

Key Changes:

Region Model Enrichment: Added latitude, longitude, and bounding box fields to the Region model, along with a self-referential parent field for hierarchical grouping.
New map_discovery App:
GET /api/map/regions/: Feed for placing pins/markers with content counts.
GET /api/map/discover/: Primary endpoint for map viewport (bounding box) queries.
GET /api/map/regions/<id>/content/: Unified, paginated content feed for a specific region.
Content Model Updates:
Added direct region support to Story models.
Migrated CulturalContent to use proper Region ForeignKeys.
Data Seeding: Added python manage.py seed_region_geodata command to populate coordinates for ~30 common cultural/geographic regions.
Testing: Comprehensive test suite with 41 passing tests covering all new logic and edge cases.
Frontend Compatibility Guarantee:

Zero Breaking Changes: All existing API responses (Recipes, Stories, Cultural Content) continue to return region data as string names (e.g., "Anatolia") rather than database IDs.
Existing frontend consumers do not need any updates to maintain current functionality.


laternis (commented):
Solid foundation for the map discovery work — bbox endpoints are well-structured and the backward-compat shim that keeps Region as a string in existing serializers is clean. A few things to address before merge:

**Critical**
- `apps/map_discovery/views.py` — `BoundingBoxDiscoverView` does `longitude__lte=east, longitude__gte=west`. When the viewport crosses the antimeridian (`east < west`, e.g., Pacific viewports), this matches nothing and silently returns an empty result. Either reject `east < west` with a 400 in the validator, or branch to `Q(longitude__gte=west) | Q(longitude__lte=east)` when the bbox wraps. No test covers this case.
- `apps/map_discovery/views.py` — `RegionContentView.get()` materializes all three content types into Python lists, merges and sorts in memory, *then* paginates. ORM `LIMIT` is never applied, so a region with thousands of recipes will pull them all per request and the `MapPageNumberPagination.max_page_size` is effectively ignored. Apply `[:N]` per type before serializing, or paginate at the queryset level.

**Important**
- `apps/cultural_content/migrations/0003_backfill_cultural_content_region_fk.py` — `filter(region=None)` should be `region__isnull=True` for FK fields (Django's coercion to IS NULL on FK exact-match isn't reliable across versions). Per-row `obj.save()` is fine for current volumes but `bulk_update(..., ['region'])` is the standard data-migration pattern. Also worth pairing with a reverse step.
- `apps/cultural_content/serializers.py` — `get_region()` now reads `obj.region.name`, which adds a SELECT per row unless the CulturalContent list/retrieve viewset uses `select_related('region')`. The PR doesn't show the viewset getting that hint — please confirm it's there or add it; otherwise this is an N+1 regression on an existing endpoint.

**Minor (non-blocking)**
- `apps/map_discovery/views.py` — `authentication_classes = []` on all four views disables JWT parsing entirely, so even authenticated requests won't have `request.user` populated. For public read endpoints prefer keeping default authentication and only setting `permission_classes = [AllowAny]`, so future per-user instrumentation/audit has access to `request.user`.

Once the antimeridian handling and the unbounded `RegionContentView` pagination are addressed (and the migration FK-filter / select_related items confirmed), I'm happy to approve.

laternis (approved):
LGTM — all feedback addressed cleanly with tests for each fix.

PR #458, opened 2026-05-08
feat: refactor to story-centric data model #379

### LINKED ISSUE #379: #M5-13 — Story-centric data model
## Description
Stories become the primary unit; recipes are linked supporting elements. Migrate existing content; adjust foreign keys so a recipe can belong to a story and a story can reference multiple recipes.

## Motivation / Context
Core pivot from recipe-first to story-first identified in the LGTM review.

## Deliverables
- Updated Story / Recipe relations
- Data migration for existing content

## Acceptance Criteria

## References
- Draft requirement D.1.1

roboticrustacean:
Transitioned from recipe-first to story-centric architecture. Implemented Many-to-Many relationship between stories and recipes with ordering. Added summary and updated_at fields to stories. Maintained backward compatibility for single linked_recipe input/output. Updated search and map discovery services to support M2M links. Verified with comprehensive test suite. 
Closes #379


laternis (commented):
Migration shape (0005 schema → 0006 data → 0007 drop legacy) is the right approach, and the `linked_recipe` backward-compat on serializer input/output is a helpful bridge. Some prefetch and migration-correctness issues to fix:

**Critical**
- `apps/search/views.py` `_serialize_story` — `story.recipe_links.all().exists()` followed by `story.recipe_links.all()[0]` is two DB hits per story even though `recipe_links__recipe__region` is prefetched. Materialize once:
  ```python
  links = list(story.recipe_links.all())
  first = links[0] if links else None
  ```
- `apps/stories/serializers.py` — `get_recipe_title` calls `obj.recipe_links.select_related('recipe').first()`. Adding `.select_related()` on a prefetched M2M bypasses the prefetch cache and refires per story. `get_linked_recipe` and `get_region_name` walk `recipe_links` separately as well. A small helper that reads from the cache would collapse all three to zero extra queries:
  ```python
  def _first_link(self, obj):
      links = list(obj.recipe_links.all())
      return links[0] if links else None
  ```

**Important**
- `apps/stories/migrations/0006_migrate_linked_recipes.py` — `backwards()` uses `story.recipe_links.first()` against the live manager instead of the historical model from `apps.get_model('stories', 'StoryRecipeLink')`. If the through model's schema ever evolves, rolling back past 0006 will use the wrong state. Mirror the `forwards` pattern with the historical model.
- `apps/stories/serializers.py` `_set_recipes` — `elif single_id:` treats `linked_recipe_id=0` as "no value" instead of validating. PKs in practice are ≥1 so this is defensive, but `elif single_id is not None:` is consistent with the `update` guard above it and fails loudly on bad input.

**Minor (non-blocking)**
- No test asserts that `linked_recipes` preserves `StoryRecipeLink.Meta.ordering = ['order', 'added_at']` on read. `assertIn` covers presence but not order — would be valuable since ordering is the whole reason the through model exists. Same for `RecipeViewSet.story_count` annotation: a quick assertion that it's non-zero after a story links a recipe would catch silent annotation drops.

Once the prefetch-bypass issues in `_serialize_story` / story serializers and the historical-model use in `0006.backwards` are addressed, I'm happy to approve.

laternis (approved):
LGTM — prefetch fixes look right, historical model used correctly in 0006.backwards.

Issue #418, opened 2026-04-30
Mobile: swap orange/white roles and update orange to #E69138


CaganIslam:
## Decision
- Swap the orange and white roles across the mobile theme — wherever orange is used, place white; wherever white is used, place orange.
- Change the orange hex to **#E69138** (hover: **#B8732D**).

Goal: a brighter, white-dominant surface treatment with orange used as accent (the inverse of the current usage).

## Scope
- Mobile only (`app/mobile/src/theme`).
- Verify dependent token aliases (`surface`, `primary`, `primarySubtle`, `primaryBorder`, `primaryTint`, `bg`, `text`, etc.) still produce a coherent UI after the swap.
- Visual regression sweep on the emulator: Home, RecipeDetail (+ Q&A panel), StoryDetail, Search, Profile/Auth flows.

## Acceptance Criteria
- [ ] Orange constant updated to #E69138
- [ ] Token role swap applied (orange ↔ white)
- [ ] Major screens visually verified on emulator with no broken contrast or unreadable text
- [ ] No raw hex literals introduced in screens — changes go through theme tokens

PR #460, opened 2026-05-09
fix(mobile): restore button and inbox contrast after palette swap (#456)

CaganIslam:
## Summary
Profile tab buttons and Inbox rows lost their visible borders after the palette swap (#418/#426). The new `primary` token equals `surface`, so any element bordered by `primary` on a `surface` card became invisible.

## Fix
- `ProfileTabScreen` — three colored action buttons (Messages green, Cultural preferences yellow, Log out red), all with `surfaceDark` border and full-width consistent layout.
- `InboxScreen` — rows now use cream `bg` instead of `surface`, with a thicker `surfaceDark` border. Time, preview, and unread preview text moved off `textMuted` so they read clearly.

## Test
- `npx tsc --noEmit` clean
- Local emulator: Profile shows three distinct buttons; Inbox rows are readable; empty state and unread badge unaffected.

Closes #456


ErenCanOzkaya (approved):
LGTM

PR #467, opened 2026-05-09
feat(backend): cultural tag moderation queue (#391)

laternis:
## Summary
- Audit fields (submitted_by/at, reviewed_by/at, rejection_reason) on EventTag, Region, and Religion via a shared CulturalModerationMixin.
- New admin endpoints under /api/moderation/cultural-tags/: list pending/approved/rejected queue with type filter, approve, reject with reason.
- Submission dedup logic: approved duplicate returns 409 with the existing record, pending duplicate returns 200 with a "queued" payload, no row is double-created.
- RegionViewSet expanded to allow authenticated POST that lands as is_approved=False; existing seeded regions backfilled to is_approved=True via data migration.

## Test plan
- [x] `python manage.py test apps.recipes apps.cultural_content -v 2` (24 new moderation tests pass; pre-existing failures from #419 seed flakiness are unchanged on main)
- [x] `python manage.py test apps.search apps.stories apps.map_discovery apps.common apps.notifications apps.users apps.messaging` (168 tests, no regressions)
- [x] new tests cover submit / queue list / approve / reject / dedup / non-admin permission paths
- [ ] manual smoke: admin login, list queue, approve a row, verify it appears in public listing

## Notes
- Moderation endpoints live in `apps/cultural_content/moderation_views.py` since the cultural_content app is the natural home for this surface; the existing per-type submission viewsets in `apps/recipes` keep handling create/dedup. Shares the `is_approved` field already in use across the moderated lookup family; this PR adds audit metadata and the admin surface.
- Region is now user-submittable (matching the issue intent of "user-submitted region/event/tradition tags"). Geographic metadata (lat/lng/bbox/parent) stays admin-only on submission via a dedicated RegionSubmissionSerializer; admins can set those fields via the Django admin or a follow-up endpoint.
- Even staff submissions land as is_approved=False on POST — staff approves later via the dedicated approve endpoint, keeping a single audit-trail path.
- DietaryTag intentionally stays on the older NamedSubmissionSerializer dedup (400 on duplicate) since it isn't a cultural tag per the issue scope.
- Ingredient/unit moderation queue (#M6-10) can reuse the same shape; the type discriminator dict in `moderation_views.CULTURAL_TAG_TYPES` is the only place to extend.
- Frontend admin UI (the issue's `area: frontend` deliverable) is covered by the frontend cohort; this PR is backend-only.

Closes #391.

PR #466, opened 2026-05-09
feat(backend): conversion subsystem (#374, #375, #376)

laternis:
## Summary
- Pure unit conversion engine covering metric, imperial, cup/spoon, and informal units (#374).
- `Ingredient.density_g_per_ml` plus seeded density table with cited sources for accurate mass<->volume (#375).
- `POST /api/convert/` public endpoint consuming the engine and density table (#376).

## Test plan
- [x] `python manage.py test apps.recipes -v 2` (36 new tests pass, the 19 errors + 3 failures match `main` and are tracked under #419 about flaky seed migrations)
- [x] new tests: conversion engine pairs, density seed idempotency, `/convert` 200 + 400 paths
- [x] `python manage.py makemigrations --check --dry-run` clean
- [ ] manual smoke: `curl -X POST .../api/convert/ -d '{"amount":"1","from_unit":"cup","to_unit":"ml"}'`

## Notes
- OpenAPI generator is not wired into the project; endpoint contract is documented in `apps/recipes/conversions/README.md`. Happy to switch to drf-spectacular in a follow-up.
- Density values are cited in `apps/recipes/conversions/references.md` (USDA FoodData Central, King Arthur weight chart, CRC Handbook). No uncited values landed.
- Seeder uses `get_or_create` and only updates `density_g_per_ml` so it cohabits cleanly with the `0005_seed_ingredients_units` migration (defensive against the UNIQUE collision noted in #419).
- Taking over from @durakahmetayberk on these three; he is staying as co-assignee.

Closes #374, closes #375, closes #376.

Issue #419, opened 2026-04-30
Recipes seed migrations break tests with UNIQUE collisions and lookup assertion mismatches


laternis:
## Description
Running the full backend test suite produces **47 ERRORs and 3 FAILures** across `apps.recipes`. The backend itself works correctly in dev/prod — the bug is that legacy tests assume an empty `Region`/`Ingredient`/`Unit` table, but two data migrations now seed those tables before tests run.

```
Ran 137 tests in 23.655s
FAILED (failures=3, errors=47)
```

`apps.users`, `apps.notifications`, `apps.messaging`, `apps.stories` are unaffected.

## Root cause
Two data migrations seed the lookup tables on every DB build (including the test DB):

- `apps/recipes/migrations/0004_seed_regions.py` — 32 regions (Aegean, Mediterranean, Italian, …)
- `apps/recipes/migrations/0005_seed_ingredients_units.py` — 45 ingredients + 12 units (Tomato, Salt, cups, tablespoons, …), all `is_approved=True`

Older tests that pre-date these migrations still create the same names by hand in `setUp`, which trips the `name` UNIQUE constraint and aborts setUp.

### Two failure modes

**1) `IntegrityError: UNIQUE constraint failed` (47 errors)**
Examples:
- `apps/recipes/tests.py:16` — `Region.objects.create(name=\"Mediterranean\")`
- `apps/recipes/tests.py` — `Unit.objects.create(name=\"pieces\", is_approved=True)`
- `apps/recipes/tests_lookup_apis.py` — `Ingredient.objects.create(name='Apple', is_approved=True)` (Apple isn't seeded, but \`cup\`/\`Tablespoon\` etc. collide with seed entries \`cups\`/\`tablespoons\` — case-insensitive collision in \`tests_custom_submission_api.py\`)

When \`setUp\` raises, every test in the class is reported as ERROR, which is why a handful of collisions inflate to 47 errors.

**2) Lookup assertion mismatches (3 failures)**
Affected:
- \`tests_lookup_apis.LookupApiTests.test_unit_lookup_returns_approved_units_for_dropdowns\`
- \`tests_lookup_apis.LookupApiTests.test_ingredient_lookup_returns_approved_ingredients_for_dropdowns\`
- \`tests_custom_submission_api.CustomSubmissionApiTests.test_authenticated_user_can_submit_a_new_unit\`

[...]

PR #468, opened 2026-05-09
fix(backend): merge conflicting recipes 0014 migrations

laternis:
## Summary
- Adds `recipes/migrations/0016_merge_20260509_1145.py` to resolve the multiple-leaf graph in `recipes` left behind by PRs #466 and #467 both branching from `0013_seed_taxonomy`.
- No DB schema change. Empty operations; deps point at both 0014 leaves so Django sees a single leaf again.

## Test plan
- [x] `python manage.py migrate --plan` runs cleanly (previously errored with "Conflicting migrations detected; multiple leaf nodes")
- [x] Generated via `python manage.py makemigrations --merge`, no manual edits

## Notes
- Background: PRs #466 (conversion subsystem) and #467 (cultural tag moderation) ran in parallel from `origin/main` and both landed migration `0014_*` on top of `0013_seed_taxonomy`. They were marked MERGEABLE because text didn't conflict, but the migration graph collided. Both merged before this was caught, breaking `migrate` on `main`.
- For future parallel backend work in the same Django app, second PR should rebase + rename its migration before merging, or we add a CI check that fails on multi-leaf graphs.

Closes nothing (no issue filed; see #419 for related migration tech debt).


roboticrustacean (approved):
thanks for the fix

PR #465, opened 2026-05-09
feat(backend): domain-aware search query parsing (#389)

laternis:
## Summary
- Lookup-driven query parser extracts region / event / diet / religion facets from natural-language search queries.
- `GlobalSearchView` injects parsed facets into the existing filter pipeline; cleaned residual query feeds title/body matching.
- Parsed facets echoed in the response for client debug, and the personalization scorer from #463 gets a small region/event weight boost when those facets came from the parser.

## Test plan
- [x] `python manage.py test apps.search -v 2` (45 tests, all green)
- [x] `python manage.py test apps.recipes` (same 3F/19E baseline as `origin/main`, see Notes)
- [x] new tests: 19 parser unit cases + 8 endpoint-level decomposition cases
- [ ] manual smoke: `curl ".../api/search/?q=Balkan%20wedding%20dishes"`

## Notes
- Parser is lookup-driven, not LLM-driven, to keep latency bounded and behavior predictable. Multi-word region names (e.g. "Eastern European", "Black Sea") handled via longest-match-first.
- Explicit query params still win over parsed facets; we never override user intent.
- Parser tests pass a fixed in-memory lookup map so the matching logic stays DB-free.
- `apps.recipes` shows the same 3 failures + 19 errors on this branch as on `origin/main`. They are the seed-migration flakiness tracked in #419 (Ingredient/Unit unique-constraint and lookup-ordering issues), unrelated to this change. Per the prompt I am not papering over them.

Closes #389.

PR #470, opened 2026-05-09
fix(backend): unbreak recipes test suite from seed migration collisions (#419)

### LINKED ISSUE #419: Recipes seed migrations break tests with UNIQUE collisions and lookup assertion mismatches

laternis:
## Summary
- Renames legacy fixtures in `apps/recipes/tests_permissions.py` and `tests_custom_submission_api.py` so they no longer collide with rows seeded by `0004_seed_regions` and `0005_seed_ingredients_units`.
- Tightens the three lookup-endpoint assertions in `tests_lookup_apis.py` to scope by created IDs instead of asserting full equality against a seeded universe.

## Test plan
- [x] `python manage.py test apps.recipes -v 2` (was 19 errors + 3 failures, now 0/0 across 140 tests)
- [x] `python manage.py test` (full suite green: 371/371)

## Notes
- Picked option 1 from the issue (smallest blast radius). Seeds are untouched; prod and test continue to share the same migration path.
- Lookup assertions still verify the fixture rows' full shape; just filtered to the IDs the test created so they ignore seeded rows.
- Three follow-on PRs from today (#465, #466, #467) referenced this issue when reporting baseline noise. After this lands, future backend PRs should land with a clean suite.

Closes #419.

### DIFF
--- a/app/backend/apps/recipes/tests_custom_submission_api.py
+++ b/app/backend/apps/recipes/tests_custom_submission_api.py
@@ -29,10 +29,10 @@ class CustomSubmissionApiTests(APITestCase):
         self.assertFalse(response.data['is_approved'])
 
     def test_authenticated_user_can_submit_a_new_unit(self):
-        response = self.client.post(self.unit_list_url, {'name': '  Pinch '})
+        response = self.client.post(self.unit_list_url, {'name': '  TestUnit-zeta '})
 
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
-        self.assertEqual(response.data['name'], 'Pinch')
+        self.assertEqual(response.data['name'], 'TestUnit-zeta')
         self.assertFalse(response.data['is_approved'])
 
     def test_ingredient_submission_rejects_blank_names(self):
@@ -48,9 +48,9 @@ class CustomSubmissionApiTests(APITestCase):
         self.assertEqual(response.data['name'][0], 'This field may not be blank.')
 
     def test_ingredient_submission_rejects_case_insensitive_duplicates(self):
-        Ingredient.objects.create(name='Salt', is_approved=True)
+        Ingredient.objects.create(name='TestIngredient-omega', is_approved=True)
 
-        response = self.client.post(self.ingredient_list_url, {'name': '  salt  '})
+        response = self.client.post(self.ingredient_list_url, {'name': '  testingredient-omega  '})
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertEqual(response.data['name'][0], 'An ingredient with this name already exists.')
--- a/app/backend/apps/recipes/tests_lookup_apis.py
+++ b/app/backend/apps/recipes/tests_lookup_apis.py
@@ -22,8 +22,12 @@ class LookupApiTests(APITestCase):
         response = self.client.get(self.ingredient_list_url)
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+        # Scope to fixtures created here. The endpoint also returns rows seeded
+        # by 0005_seed_ingredients_units; assertion stays exact for our IDs.
+        created_ids = {self.apple.id, self.banana.id}
+        filtered = [row for row in response.data if row['id'] in created_ids]
         self.assertEqual(
-            response.data,
+            filtered,
             [
                 {'id': self.apple.id, 'name': 'Apple'},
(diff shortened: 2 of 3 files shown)


roboticrustacean (approved):
thanks for the fix. merging now

PR #473, opened 2026-05-09
chore(backend): trace Lab 9 backend acceptance tests + fill scenario gaps (#472)

laternis:
## Summary
- Each of the 5 TC_API_* tests committed in Lab 9 (AUTH_001, REC_002, ING_003, STORY_004, QA_005) now carries an in-source docstring with the Lab 9 ID, designer, and requirement IDs.
- Where current coverage was partial (JWT decode + claim assertions, post-403 unchanged-recipe verification, full ingredient pending->approve flow, bidirectional story-recipe back-link, qa_enabled toggle-and-retry leg), the existing tests have been extended.
- New \`docs/lab9-acceptance-test-traceability.md\` maps each TC_API_* to its real file/class/method so the Final Milestone Report can cite Lab 9 with a verifiable link.

## Test plan
- [x] \`python manage.py test apps.users apps.recipes apps.stories -v 2\` green (211 tests, all pass)
- [x] \`grep -rn "TC_API_" app/backend/apps/ docs/\` returns all 5 IDs across test docstrings + traceability doc

## Notes
- Lab 9's hypothetical file paths (\`apps/ingredients/...\`, \`apps/comments/...\`, \`apps/users/tests/...\`) don't match the actual app layout. The traceability doc explicitly records the divergence and points at the real paths. We did not rename existing test methods to match Lab 9's literal strings, that would be churn for no value. The docstring trace is what makes the commitment verifiable.
- Lab 9 also cited a hypothetical \`linkedStories[]\` field on the Recipe API response for STORY_004. The current Recipe serializer (post-#458) exposes the back-link as \`story_count\` plus the model-level reverse manager \`Recipe.linked_stories\`. Both are asserted in the test.
- Frontend Cypress tests (TC_UI_006 -> 009) are not in scope; they belong to the frontend cohort.
- CI gating for these tests is tracked separately under #469.

Closes #472.

PR #476, opened 2026-05-09
fix(devops): pass CORS_ALLOWED_ORIGINS to backend and force-recreate on deploy

laternis:
## Summary
- Add `CORS_ALLOWED_ORIGINS: ${CORS_ALLOWED_ORIGINS:-https://genipe.app,https://www.genipe.app}` to the backend service's `environment` block in `docker-compose.yml`. Was missing entirely; backend was crashlooping with `corsheaders.E013: Origin '' in CORS_ALLOWED_ORIGINS is missing scheme or netloc` whenever the container actually got recreated against the current image, because `os.getenv('CORS_ALLOWED_ORIGINS', '').split(',')` resolved to `['']`.
- Append `CORS_ALLOWED_ORIGINS=https://genipe.app,https://www.genipe.app` to the root `.env` on first deploy if missing (idempotent, mirrors the existing `ALLOWED_HOSTS` and `REACT_APP_API_URL` lines in `deploy-web.yml`).
- Switch the deploy workflow's `docker compose ... up -d` to `up -d --force-recreate`. Today's outage came from the web container being created weeks ago without the `docker-compose.prod.yml` overlay, so 443 was never bound. Compose's normal diff doesn't recreate that container even when ports or volumes are added to the overlay later, which is the real root cause behind 5 consecutive failed deploys today.

## Test plan
- [x] Applied the equivalent edits live on prod (sed-injected the env line, force-recreated backend) — backend now booted past the system check, gunicorn listening on `:8000`, and external smoke `https://genipe.app/`, `/api/recipes/`, `/api/search/?q=mercimek` all return 200.
- [x] `docker compose -f docker-compose.yml -f docker-compose.prod.yml config` shows the merged `web` service binds 443 and mounts `/etc/letsencrypt` + `ops/nginx-prod.conf`. Force-recreate of `web` produced a container with both 80 and 443 host-bound, HTTP-to-HTTPS redirect 301 working.
- [ ] After this merges and the auto-deploy runs, the workflow itself should reach the smoke-test step green for the first time today.

## Notes
- Live prod is healthy as of this PR open. The `sed` patch on the server's `docker-compose.yml` will be overwritten by the next deploy's `git reset --hard origin/main`, but the same change is in this PR, so the next deploy lands an equivalent config — no flapping risk if this gets merged before someone triggers an unrelated deploy.
- Considered also filtering empty strings from the CORS split inside `settings.py` for defense in depth (`[o for o in os.getenv(...).split(',') if o]`). Held off to keep the PR scoped to the deploy/runtime gap; happy to add it as a follow-up if reviewer prefers.
- The deploy workflow's smoke test step has been failing on every push today (#465, #466, #467, #468, #470, #473) because of the unbound 443. Without `--force-recreate`, fixing the underlying compose config alone would not have rescued the running container.

### DIFF
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -47,6 +47,8 @@ jobs:
               echo 'ALLOWED_HOSTS=genipe.app,www.genipe.app,45.76.34.210,localhost,backend' >> .env
             grep -q '^REACT_APP_API_URL=' .env || \
               echo 'REACT_APP_API_URL=https://genipe.app' >> .env
+            grep -q '^CORS_ALLOWED_ORIGINS=' .env || \
+              echo 'CORS_ALLOWED_ORIGINS=https://genipe.app,https://www.genipe.app' >> .env
 
             # Build images first. If this fails, nothing on the running host
             # is touched — legacy stack keeps serving.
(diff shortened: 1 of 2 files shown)


roboticrustacean (approved):
thanks for the fix

PR #489, opened 2026-05-09
fix(mobile): unwrap paginated /api/stories/ via fetchStoriesList helper (#488)

CaganIslam:
## Summary
Fix for #488. Stories were empty on Home and UserProfile because both screens called `apiGetJson('/api/stories/')` and assumed an array, but the backend returns DRF-paginated `{count, results}`.

## Files
- `services/storyService.ts` — new `fetchStoriesList()` that handles both raw arrays and `{results}` envelopes
- `screens/HomeScreen.tsx` — call `fetchStoriesList()`, drop the inline unwrap
- `screens/UserProfileScreen.tsx` — same swap

## Test
- `npx tsc --noEmit` clean
- Local docker stack from `LOCAL_DEV.md`: Home shows the 16 stories; UserProfile shows the user's stories.

## Notes
- Out of scope: lazy-loading via `next` cursor, and making `apiGetJson` itself pagination-aware (would silently drop `count/next/previous` metadata).

Closes #488

### DIFF
--- a/app/mobile/src/screens/HomeScreen.tsx
+++ b/app/mobile/src/screens/HomeScreen.tsx
@@ -4,7 +4,7 @@ import { SafeAreaView } from 'react-native-safe-area-context';
 import { useEffect, useMemo, useState } from 'react';
 import { shadows, tokens } from '../theme';
 import { fetchRecipesList } from '../services/recipeService';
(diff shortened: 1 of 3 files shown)


ErenCanOzkaya (approved):
LGTM

Issue #503, opened 2026-05-09
Backend: POST /api/convert/ returns 401 despite AllowAny permission


CaganIslam:
## Description
The `ConvertView` in `app/backend/apps/recipes/views.py` declares `permission_classes = [permissions.AllowAny]`, but live calls without an Authorization header return `401 Unauthorized`. The view's behaviour does not match its declared contract.

## Repro
```
curl -i -X POST http://localhost/api/convert/ \
  -H "Content-Type: application/json" \
  -d '{"amount":"100","from_unit":"g","to_unit":"oz"}'
```
→ `HTTP 401 Unauthorized` with `{"detail": "Authentication credentials were not provided or are invalid."}`

## Expected
Anonymous request should succeed (or, if auth is intentionally required, the view's `permission_classes` should reflect that and the issue body of #376 + the docstring should be updated).

## Suspected cause
Likely a `DEFAULT_PERMISSION_CLASSES` set to `IsAuthenticated` at the DRF level overriding individual view declarations in some path. The `IngredientViewSet.substitutes` action handles the same trap with a custom `get_permissions` override — see if `ConvertView` needs similar treatment.

## Came from
Discovered while implementing mobile #408 (swap to backend `/convert`). Mobile is unblocked because logged-in users always send a JWT, but anonymous web visitors using the convert toggle on `RecipeDetailPage` will hit the same 401.

## Acceptance
- [ ] Anonymous `POST /api/convert/` returns 200 with the conversion result
- [ ] Or, view + spec updated to make auth requirement explicit

PR #510, opened 2026-05-09
feat(mobile): swap client-side unit conversion to backend /convert API (#408)

### LINKED ISSUE #408: Mobile: swap client-side unit conversion to backend /convert API
## Description
`app/mobile/src/utils/unitConversion.ts` (added in #378) does conversion locally with a hardcoded factor table. Once `/convert` API (#376) ships, mobile should call it instead so conversion logic lives in one place and stays consistent with web.

## Deliverables
- Replace `convertIngredient(amount, unitName)` body with a call to `POST /api/convert/` (`{ amount, fromUnit, toUnit, ingredientId? }`).
- Keep the same `ConvertedAmount | null` return shape so `RecipeDetailScreen` doesn't change.
- Cache responses per `(amount, fromUnit, toUnit)` key for the lifetime of the screen — avoid re-hitting the API on every re-render.
- Loading state on the toggle while a batch of conversions is in flight.

## Acceptance Criteria
- [ ] Toggle uses `/convert` for all mapped units; no local factors remain.
- [ ] Network failures fall back to original (unchanged) display, not crash.
- [ ] No visible flicker/jank when toggling (cache works).
- [ ] Manual test against local backend confirms parity with web `/convert` results.

## References
- Depends on: #376 (`/convert` API)
- Touches: `app/mobile/src/utils/unitConversion.ts`, `app/mobile/src/screens/RecipeDetailScreen.tsx`
- Introduced by: #378 / PR #407

CaganIslam:
## Summary
Closes #408. Replaces the local hardcoded factor table in `unitConversion.ts` with calls to `POST /api/convert/`. The recipe detail "Converted" toggle now shows server-computed values, with an in-memory cache so toggling back and forth doesn't re-hit the API.

## Files
- `services/unitConversionService.ts` — `fetchConversion(amount, fromUnit, toUnit, ingredientId?)` with module-level cache keyed by the input tuple
- `utils/unitConversion.ts` — kept the source-to-target unit mapping (`targetUnitFor`); removed the math
- `screens/RecipeDetailScreen.tsx` — `useEffect` batches a request per ingredient when the user flips to "Converted"; shows "Converting…" while in flight; falls back to the original amount on network failure
- `screens/HomeScreen.tsx` + `navigation/PublicStackNavigator.tsx` — small fixes for merge resolution leftovers picked up while on this branch

## Test
- `npx tsc --noEmit` clean
- Local docker stack: logged in as `ayse@example.com`, opened Greek Dolmadakia. Toggling "Converted" replaces 200 grams Rice with `7.0548 oz` and 80 ml Olive Oil with `2.6667 fl oz`. Toggling again is instant (cache hit). Pieces / bunch / pinch ingredients keep their original display because no target unit is mapped — that's the intended scope.

## Notes
- Anonymous users see the original values without conversion because `POST /api/convert/` returns 401 anon despite the view declaring `AllowAny`. Tracked in #503 (backend). Mobile is unblocked since logged-in users always send a JWT.
- Pieces/bunch/pinch and other counted units have no target unit by design — they pass through.

Closes #408

### DIFF
--- a/app/mobile/src/navigation/PublicStackNavigator.tsx
+++ b/app/mobile/src/navigation/PublicStackNavigator.tsx
@@ -100,6 +100,8 @@ export function PublicStackNavigator() {
         name="MapDiscovery"
         component={MapDiscoveryScreen}
         options={{ title: 'Discover by region' }}
+      />
+      <Stack.Screen
         name="Explore"
         component={ExploreScreen}
         options={{ title: 'Explore' }}
--- a/app/mobile/src/screens/HomeScreen.tsx
+++ b/app/mobile/src/screens/HomeScreen.tsx
@@ -104,6 +104,9 @@ export default function HomeScreen({ navigation }: Props) {
             <Text style={styles.mapSubtitle}>Tap a pin on the map to focus a region</Text>
           </View>
           <Text style={styles.mapArrow}>→</Text>
+        </Pressable>
+
+        <Pressable
           onPress={() => navigation.navigate('Explore')}
           style={({ pressed }) => [styles.exploreEntry, pressed && styles.exploreEntryPressed]}
           accessibilityRole="button"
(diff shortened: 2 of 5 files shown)


mustafaocakxyz (approved):
Nice implementation, thank you.

PR #538, opened 2026-05-09
fix(mobile): restore comment submit button and Q&A type toggle visibility (#493)

CaganIslam:
## Summary
Closes #493. The comment submit button and the Question/Comment toggle on the recipe Q&A panel were invisible after the palette swap (#418/#426). Both used `tokens.colors.primary` (orange) on a `surface` (orange) card — same regression family as #456.

## Files
- `components/recipe/RecipeCommentsSection.tsx` — submit button now `accentGreen` fill with `surfaceDark` outline and `textOnDark` text. Active Q&A type toggle (Question / Comment) uses the same accent so the selected state reads.

## Test
- `npx tsc --noEmit` clean
- Local docker stack: typed a comment on a recipe — the green "Post" button shows up; toggling between Question and Comment makes the selected pill clearly visible. Disabled state still dims the button when the input is empty.

Closes #493


ErenCanOzkaya (approved):
LGTM

Issue #364, opened 2026-04-18
M6-11 — Auto-save draft & restore


ErenCanOzkaya:
## Description
In-progress edits are auto-saved as drafts and offered for restoration on return to the editor.

## Deliverables
- Draft model + endpoints
- Web + mobile auto-save + restore flows

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 4.4.3

## Notes

PR #547, opened 2026-05-10
feat: implement auto-save draft model, endpoints, and cleanup signals

roboticrustacean:
Implemented a robust backend system for auto-saving in-progress edits to prevent data loss.

New drafts app: Generic model with JSONField to store editor state for any entity type.
POST /api/drafts/ handles both creation and periodic auto-saves using update_or_create.
Integrated Django signals to delete drafts automatically when a Story or Recipe is successfully saved/published.
Unit test coverage for the draft lifecycle, including uniqueness constraints and signal cleanup.

#364 's backend side is handled by this pr.


laternis (commented):
Solid foundation for the drafts subsystem. The two partial unique constraints on the model read well, the upsert via `update_or_create` is the right shape for periodic auto-saves, and the test coverage spans model, API, and signal layers. Two things to address before merge.

## Critical

`apps/drafts/signals.py:7-20` (and the symmetric recipe receiver at lines 22-35) deletes every `target_id__isnull=True` draft on every `post_save`, including updates. Scenario: a user has Story A saved, starts a brand-new Story B as a draft with `target_id=None`, then opens Story A and saves a small edit. Story A's `post_save` fires, the second `Draft.objects.filter(... target_id__isnull=True)` block wipes the unrelated Story B draft.

The cleanup of "new" drafts should only run on creation:

```python
@receiver(post_save, sender=Story)
def delete_story_draft(sender, instance, created, **kwargs):
    Draft.objects.filter(
        user=instance.author,
        target_type='story',
        target_id=instance.public_id,
    ).delete()
    if created:
        Draft.objects.filter(
            user=instance.author,
            target_type='story',
            target_id__isnull=True,
        ).delete()
```

Existing tests don't catch this because `test_story_update_cleans_draft` uses a draft tied to the story's `public_id`, not a separate "new" draft. A regression test of the form "story update preserves an unrelated `target_id=None` draft for the same user" would lock the fix in.

## Important

`apps/drafts/views.py` `create()` reads `target_type`, `target_id`, and `data` directly from `request.data`, bypassing the serializer. Two consequences:

1. `target_type` is a free-form string. A client can post `target_type='foo'` and create drafts that no signal will ever clean up. Adding `choices=[('story', 'Story'), ('recipe', 'Recipe')]` on the model field (or at least a whitelist check in the view) keeps the surface bounded.
2. `target_id` is not length-checked against the model's `max_length=26`, and `data` has no size cap, so a client can write arbitrarily large blobs into the JSONField.

Routing the input through `DraftSerializer` (with `target_type` validation added) and doing the `update_or_create` on `serializer.validated_data` addresses both without changing the upsert behavior.

Once these two are addressed I'm happy to approve.

laternis (approved):
LGTM

PR #562, opened 2026-05-10
fix(mobile): show region as a clickable pill on Recipe and Story detail (#545, #561)

CaganIslam:
## Summary
Closes #545 and closes #561. The recipe and story detail screens used to render `recipe.region` / `story.region` directly, but the backend exposes those fields as the FK pk (an integer like `3`) and surfaces the human label in `region_name`. That's where the stray "3" was coming from on Trabzon Butter Kuymak. This PR normalizes the field to the friendly name and turns it into a tappable pill that opens Search filtered by that region — same affordance the map's CTA uses.

## Files
- `services/recipeService.ts` — `normalizeRecipeDetail` resolves `region` from `region_name` (or a nested `{name}` object)
- `services/storyService.ts` — `normalizeStoryDetail` does the same for stories; surfaces `region` on the normalized payload
- `types/story.ts` — `StoryDetail` now declares the optional `region: string`
- `screens/RecipeDetailScreen.tsx` — region rendered as a pill (cream fill, surfaceDark border); tap navigates to `Search` with `?region=<name>`
- `screens/StoryDetailScreen.tsx` — same pill mirroring the recipe behaviour

## Test
- `npx tsc --noEmit` clean
- Local docker stack: opened Trabzon Butter Kuymak — title now shows "Black Sea" pill instead of the stray "3". Tapping it opens Search filtered by Black Sea. Stories with a region get the same pill. Recipes/stories without a region don't render an empty pill.

Closes #545
Closes #561

### DIFF
--- a/app/mobile/src/screens/RecipeDetailScreen.tsx
+++ b/app/mobile/src/screens/RecipeDetailScreen.tsx
@@ -148,7 +148,19 @@ export default function RecipeDetailScreen({ route, navigation }: Props) {
           <Text style={styles.title} accessibilityRole="header">
             {recipe.title}
           </Text>
-          {recipe.region ? <Text style={styles.meta}>{recipe.region}</Text> : null}
+          {recipe.region ? (
+            <Pressable
+              onPress={() =>
+                navigation.navigate('Search', { region: recipe.region as string })
+              }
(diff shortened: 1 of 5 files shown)


ErenCanOzkaya (approved):
Nice job, LGTM

PR #567, opened 2026-05-10
fix(frontend): explore page, search filter accordions, map content loading

ErenCanOzkaya:
## Summary

  - **Explore page**: The page was broken because `exploreService` called a
    non-existent `/api/explore/events/` endpoint. Fixed to use the existing
    `/api/recommendations/?surface=explore` endpoint and adapt the flat response
    into the event-rail shape the UI expects.

  - **Search filters**: Replaced the three always-visible filter boxes (Dietary
    Tags, Event Tags, Ingredients) with collapsible accordion panels. Panels are
    closed by default and show an active-selection count badge when filters are
    selected.

  - **Map page**: Fixed `lat`/`lng` → `latitude`/`longitude` field name mismatch,
    wired up `fetchMapRegionContent` to load recipes and stories per region,
    added recipe/story counts in the panel header, and switched to the CARTO
    light tile layer for a cleaner look.
  
  - **recipeService / storyService**: Handle paginated API responses by reading
    `response.data.results` with a fallback to `response.data`.

  ## Test plan 

  - [ ] Navigate to `/explore` while logged in — recipes and stories load in two rails
  - [ ] Search for a term, open/close each accordion panel, select filters and verify badge counts update
  - [ ] Open the Map page, click different regions and verify recipe/story lists load correctly
  - [ ] Confirm recipe list and story list pages still render without regressions

### DIFF
--- a/app/frontend/src/pages/MapPage.css
+++ b/app/frontend/src/pages/MapPage.css
@@ -72,6 +72,14 @@
   color: var(--color-text);
 }
 
+.map-panel-counts {
+  display: flex;
+  gap: 1rem;
+  font-size: 0.875rem;
+  color: var(--color-text-muted);
+  font-weight: 600;
+}
+
 .map-panel-desc {
   font-size: 0.9375rem;
   color: var(--color-text-muted);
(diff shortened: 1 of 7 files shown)


CaganIslam (approved):
Tests pass (257/257). Explore endpoint fix, accordion filters and map field name corrections all look solid. LGTM

PR #569, opened 2026-05-10
fix(backend): allow anonymous POST /api/convert/ (#503)

### LINKED ISSUE #503: Backend: POST /api/convert/ returns 401 despite AllowAny permission

laternis:
## Summary
- Root cause: the custom `JWTAuthenticationMiddleware` (apps/common/middleware.py) blocks every unsafe method under `/api/` for anonymous requests before DRF gets a chance to resolve view-level `permission_classes`, so `ConvertView`'s `permission_classes = [AllowAny]` was never consulted.
- Added `/api/convert/` to `EXEMPT_POST_PATHS` so anonymous unit conversions go through.
- Set `authentication_classes = []` on `ConvertView` and updated its docstring so the public contract is explicit at the DRF layer too. View-only changes do not fix the bug on their own (middleware short-circuits first), but they document intent.
- Added a regression test that uses `django.test.Client` directly. The existing `APITestCase` suite was passing against the broken middleware because DRF's `ForceAuthClientHandler` always sets `request._force_auth_user` (even to `None`), which makes the middleware return early at the `hasattr(...)` check and skip its 401 branch. Plain `Client` reproduces the production path.

## Test plan
- [x] `python manage.py test apps.recipes` (205 tests, all green)
- [x] `python manage.py test apps.common` (21 tests, all green)
- [x] Reverted the middleware exemption locally and confirmed the new `test_convert_allows_anonymous` fails with 401, then re-applied the fix and confirmed it goes back to 200.
- [ ] Manual smoke against the dev server: `curl -X POST /api/convert/` without an Authorization header should return 200 with the converted amount.

## Notes
- View-local + middleware-local fix only. `REST_FRAMEWORK` defaults are untouched, so every other endpoint keeps its existing auth posture.
- Out of scope per the issue: `IngredientViewSet.substitutes` is auth-required by design and stays as is.

Closes #503.

### DIFF
--- a/app/backend/apps/common/middleware.py
+++ b/app/backend/apps/common/middleware.py
@@ -19,10 +19,14 @@ class JWTAuthenticationMiddleware(MiddlewareMixin):
     where a view might accidentally omit permission_classes.
     """
     
+    # Paths that are public for unsafe methods (POST/PUT/PATCH/DELETE).
+    # The middleware otherwise blocks anonymous unsafe requests to /api/* before
+    # DRF gets a chance to resolve view-level permission_classes (#503).
     EXEMPT_POST_PATHS = [
         '/api/auth/register/',
         '/api/auth/login/',
         '/api/auth/refresh/',
+        '/api/convert/',
     ]
 
     def process_request(self, request):
--- a/app/backend/apps/recipes/tests_convert_api.py
+++ b/app/backend/apps/recipes/tests_convert_api.py
@@ -1,6 +1,8 @@
 """Tests for POST /api/convert/."""
+import json
 from decimal import Decimal
 
+from django.test import Client
 from django.urls import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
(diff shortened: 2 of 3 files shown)


roboticrustacean (approved):
lgtm

PR #437, opened 2026-05-02
feat(frontend): ingredient check-off, substitution UI, and shopping list (#372, #370, #373)

Daglar1500:
## Summary
- **#372 Ingredient check-off**: checkboxes on each ingredient; checked items are faded/struck through; persisted via `POST/DELETE /api/recipes/:id/check-ingredient/` (mock: local state)
- **#370 Substitution UI**: "Sub" button on unchecked ingredients opens an inline panel with substitute options tagged by match type (Flavor Match / Texture Match); applying a sub replaces the ingredient display
- **#373 Dynamic shopping list**: collapsible panel below ingredients showing only unchecked items with applied substitutions; updates live; "Copy" button copies to clipboard
- Also includes unit toggle (#377) — **this PR supersedes `feat/frontend/unit-toggle-ui`**

## Test plan
- [ ] Log in (mock), visit `/recipes/1`
- [ ] Check an ingredient — row fades, disappears from shopping list
- [ ] Click "Sub" on unchecked ingredient — panel opens with options and match chips
- [ ] Apply a substitution — ingredient name shows strikethrough + new name
- [ ] Shopping list button appears; shows remaining unchecked items with substitutions applied
- [ ] Copy button copies list to clipboard

Issue #373, opened 2026-04-18
#M5-07 — Dynamic shopping list


laternis:
## Description
Shopping list updates live as substitutions and check-offs change. Backend persists list state; clients render and mutate.

## Deliverables
- Shopping list model + endpoints
- Web + mobile list screens
- Live update on substitute / check-off change

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.6.16
- Depends on #M5-03, #M5-06

durakahmetayberk:
abandoned the issue

Issue #372, opened 2026-04-18
#M5-06 — Available-ingredient check-off (web + mobile)


laternis:
## Description
On the recipe page, users can mark ingredients they already have. Only unchecked items feed the substitution flow and shopping list.

## Motivation / Context
Reduces noise in substitution suggestions and sets up shopping-list flow. Server-persisted to keep multi-device state consistent.

## Deliverables
- Check-off UI on ingredient list (web + mobile)
- Server-persisted check state per `(user, recipe, ingredient)` — single source of truth consumed by #M5-07 shopping list (no per-device local state to avoid multi-device desync)
- Backend endpoints to toggle and read check state

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirement 3.6.15

Issue #371, opened 2026-04-18
#M5-05 — Substitution UI (mobile)


laternis:
## Description
Equivalent substitution flow on mobile.

## Motivation / Context
Mobile parity for substitution feature.

## Deliverables
- Native substitution flow on recipe screen

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Requirements 3.6.10, 3.6.11
- Depends on #M5-03

PR #570, opened 2026-05-10
feat(backend): ingredient check-off model and endpoints (#529)

laternis:
## Summary
- New `IngredientCheckOff(user, recipe, ingredient, checked_at)` model with a unique constraint on `(user, recipe, ingredient)` and a `(user, recipe)` index.
- `GET/POST /api/recipes/<id>/checked-ingredients/`, idempotent toggle, `IsAuthenticated`. Recipe id accepts numeric pk or ULID public_id, mirroring `RecipeViewSet`.
- Migration `0019_ingredient_checkoff_529`.
- 10 tests covering auth, GET/POST happy path, idempotence, uncheck, cross-user isolation, ingredient-not-in-recipe (400), missing recipe (404), invalid body (400).

### API contract (consumed by #372 mobile, #437 web follow-up)

```
GET  /api/recipes/<id>/checked-ingredients/
  Auth: required (IsAuthenticated)
  200: [<ingredient_id>, ...]

POST /api/recipes/<id>/checked-ingredients/
  Auth: required
  Body: { "ingredient_id": <int>, "checked": <bool> }
  200: [<ingredient_id>, ...]   # canonical state after the toggle
  400: ingredient not in recipe, or invalid body
  404: recipe not found
```

`<id>` accepts the numeric pk or the ULID `public_id`. POST is idempotent: repeated `checked: true` does not create duplicate rows; `checked: false` against an unchecked ingredient is a no-op. Both POST variants return the full canonical list of checked ingredient ids so the client can reconcile without re-fetching.

## Test plan
- [x] `python manage.py test apps.recipes` (214 tests, all green)
- [x] `python manage.py test apps.users` (52 tests, all green)
- [x] `python manage.py makemigrations --check --dry-run` clean
- [x] Cross-user isolation covered by `test_users_do_not_see_each_others_checks`

## Notes
- Closes the backend half of #372. Mobile UI (#372) is in flight as a separate PR and will consume this contract.
- Web client (#437) currently uses local-only state. Migrating it onto this endpoint is a follow-up task, not this PR.
- Out of scope: shopping list (#373), substitution-aware filtering (#371), analytics/count endpoints. The surface is intentionally tight so those can compose on top.

Closes #529.


roboticrustacean (approved):
thanks. merging now

PR #572, opened 2026-05-10
fix(backend): restore IngredientSubstitution seed in canonical fixture (#542)

laternis:
## Summary
- Diagnosed: `seed_canonical` had no responsibility for the `IngredientSubstitution` graph. Migration 0010 seeds 47 pairs, but the rows are lost the moment anything wipes the table after migrate (`manage.py flush`, manual truncate, restoring an older volume), and the seed command had no way to put them back. The substitutes endpoint silently returned empty arrays whenever that happened.
- Fix: added an `ingredient_substitutions` block to `seed_canonical.json` with the same 47 pairs migration 0010 uses, and a `_seed_substitutions` step in the seed command that resolves ingredients by name and `get_or_create`s them after recipes are seeded. No new migration. No change to migration 0010. Re-runs are no-ops on healthy state; a wiped graph is restored without re-running the migration.
- Added regression tests covering: rows present after seed, anchor pair Butter to Olive Oil exists, at least 5 ingredients have non-empty substitutes, the API returns non-empty arrays, the command repopulates after a truncate, and repeat runs are idempotent.

## Test plan
- [x] `python manage.py test apps.recipes` passes (210 tests, ~36s)
- [x] `docker compose down -v && up -d` then `seed_canonical`: `curl /api/ingredients/1/substitutes/` returns `Butter -> Olive Oil` for both flavor and texture; 39 ingredients have non-empty substitutes
- [x] Truncate-and-recover path: wipe `IngredientSubstitution` while the stack is up, confirm endpoint returns empty arrays, re-run `seed_canonical`, confirm endpoint is non-empty again
- [x] Idempotent re-run: second `seed_canonical` reports `0 substitutions added (47 already present)`

## Notes
- Picked option 1 (data in the canonical fixture) over option 3 (a separate seed command). Option 3 would have created a second source of truth for the substitution graph and required documenting an extra step in `SETUP.md`. Putting the rows in `seed_canonical.json` keeps a single canonical fixture and resolves ingredients by name, so there is no FK fragility.
- The substitution data is duplicated between migration 0010's `SUBSTITUTIONS` list and the new fixture block. Migration 0010 was left untouched per the issue scope; the duplication is intentional and small (47 rows, sorted by `from`/`to`/`match_type` for stable diffs).

Closes #542.


roboticrustacean (approved):
lgtm

PR #575, opened 2026-05-10
feat(mobile): allow creating a recipe without photo or video (#550)

### LINKED ISSUE #550: Allow creating a recipe without photo or video (web + mobile)
## Description
Recipe creation currently requires a photo and/or video. Many users want to share quick recipes without media — ideas, family stories, text-only entries. The form should accept submission with no media at all.

## Repro
1. Go to Share / Create Recipe (mobile or web)
2. Fill title, description, ingredients, steps
3. Skip the image and video pickers
4. Try to submit — currently blocks (or backend rejects)

## Wanted
- Image and video are explicitly **optional** on the form
- Backend accepts a recipe without `image` / `video`
- Recipe detail screens render gracefully when both fields are null (placeholder thumb already exists, no video → existing "No video for this recipe." text)

## Files
- Mobile: `app/mobile/src/screens/RecipeCreateScreen.tsx`
- Web: `app/frontend/src/pages/RecipeCreatePage.*`
- Backend (if validation rejects): `app/backend/apps/recipes/serializers.py`

## Acceptance
- [ ] Form submits with no image and no video
- [ ] Recipe appears in feed and detail screens with placeholder visuals
- [ ] No regression on recipes that DO have media

CaganIslam:
## Summary
Closes #550 (mobile side). Recipe creation used to block submit unless a video was picked, even though the image was already optional. Some users want to share quick text-only recipes — ideas, family stories — so both media fields are now fully optional. The backend already accepts a recipe without `image` / `video` (the JSON create call doesn't carry media; the multipart PATCH only fires when at least one is selected).

## Files
- `screens/RecipeCreateScreen.tsx` — drop the `video required` validation, drop the `errors.video` field, update the helper text to call out that photo and video are optional, pass `requireSelection={false}` to the video section.
- `components/recipe/RecipeVideoSection.tsx` — render the section title as "Video (optional)" when not required, so the optional state is obvious before the user tries to submit.

## Test
- `npx tsc --noEmit` clean
- Local docker stack: opened Create Recipe, filled title + description + one ingredient, skipped both image and video pickers, hit Submit → "Recipe published!" toast, recipe shows up in the feed with the placeholder thumbnail and "No video for this recipe." text on detail.
- Regression: also created a recipe with both image and video — uploads still go through via the multipart PATCH as before.

## Out of scope
Web side of #550 (someone else's lane). The issue stays open until the web change lands.

Closes #550 (mobile)

### DIFF
--- a/app/mobile/src/components/recipe/RecipeVideoSection.tsx
+++ b/app/mobile/src/components/recipe/RecipeVideoSection.tsx
@@ -33,7 +33,7 @@ export function RecipeVideoSection({
 
   return (
     <View style={styles.section}>
-      <Text style={styles.sectionTitle}>Video</Text>
+      <Text style={styles.sectionTitle}>{requireSelection ? 'Video' : 'Video (optional)'}</Text>
 
       {hasRemoteOnly ? (
         <Text style={styles.videoHint}>
--- a/app/mobile/src/screens/RecipeCreateScreen.tsx
+++ b/app/mobile/src/screens/RecipeCreateScreen.tsx
@@ -45,14 +45,12 @@ export default function RecipeCreateScreen(_props: Props) {
     const next: {
       title?: string;
       description?: string;
-      video?: string;
       rows?: Record<string, { amount?: string; ingredient?: string; unit?: string }>;
       rowsTop?: string;
     } = {};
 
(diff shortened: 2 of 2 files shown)


ErenCanOzkaya (approved):
LGTM

PR #580, opened 2026-05-10
feat(mobile): shopping list affordance on recipe detail (#576)

CaganIslam:
## Summary
Closes #576. Web shipped a per-recipe shopping list in M5 (#373, PR #437) — a button on the recipe page that lists the ingredients you haven't checked off yet, with a one-tap copy to clipboard. Mobile had the underlying check-off state via #372 + #529 but no shopping-list affordance, so users on the phone couldn't quickly grab the "still need to buy" subset of a recipe. This PR ports the same affordance to mobile, behaviour-for-behaviour with web.

## Files
- `screens/RecipeDetailScreen.tsx` — derive `shoppingItems` from `ingredients` filtered by `checkedIds`, add a "Shopping list (N)" pill below the ingredients list, render a collapsible panel with a Copy action and the unchecked lines (or the "All ingredients are checked off!" empty state). Hidden when not authenticated or when the recipe has no ingredients.
- `package.json` / `package-lock.json` — add `expo-clipboard ~8.0.8` for the Copy action (web uses `navigator.clipboard.writeText`; the mobile equivalent is `Clipboard.setStringAsync`).

## Behaviour parity with web (PR #437)
- Same filter rule: `ingredients.filter(ri => !checkedIds.has(ri.ingredient.id))`
- Same copy format: `name — amount unit`, one per line
- Same empty state copy: "All ingredients are checked off!"
- Same auth gate: button only shows when the user is logged in

## Test
- `npx tsc --noEmit` clean
- Local docker stack (had to rebuild the backend container — the running image was stale and 404'd `/api/recipes/<id>/checked-ingredients/`; after `docker compose up -d --build backend` the endpoint serves correctly): logged in as `ayse@example.com`, opened a recipe, the "Shopping list (N)" button appeared with N matching ingredient count. Tapped it → panel opened with all lines. Tapped Copy → toast "Shopping list copied to clipboard"; pasted the list into another app and the lines matched the recipe. Checked an ingredient off → reopened the panel, the line was gone and N decreased. Checked everything off → panel showed the empty-state copy.
- Regression: inline check-off flow on the ingredient list still works exactly as before; substitutes sheet, unit toggle, and edit pill all unaffected.

## Out of scope
- The web `is-sub` (substitution) badge isn't carried over — mobile's substitution flow is a modal sheet that doesn't persist a "this row is a sub" state, so there's nothing to badge here. Filed mentally for a follow-up if substitution persistence ever lands on mobile.

Closes #576

### DIFF
--- a/app/mobile/package-lock.json
+++ b/app/mobile/package-lock.json
@@ -17,6 +17,7 @@
         "expo": "~54.0.0",
         "expo-asset": "~12.0.12",
         "expo-av": "~16.0.8",
+        "expo-clipboard": "~8.0.8",
         "expo-font": "~14.0.11",
         "expo-image-picker": "~17.0.10",
         "expo-status-bar": "~3.0.9",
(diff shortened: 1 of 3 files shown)


ErenCanOzkaya (approved):
LGTM

PR #623, opened 2026-05-10
feat(backend): story_type field on Story model and API (#565)

laternis:
## Summary
- Add `Story.story_type` (CharField, nullable, five choices: traditional, historical, family, festive, personal).
- Expose the field in list, detail, create, and update serializers.
- Add `?story_type=` filter on the stories list endpoint.

## Test plan
- [x] `python manage.py test apps.stories` green
- [x] Manual: create a story with each story_type, confirm round-trip; PATCH to a new type; list with `?story_type=traditional` returns only matches

## Notes
- Field is nullable so existing stories continue to work.
- Frontend selector is currently behind a "coming soon" gate; this PR unblocks turning it on.

Closes #565.


roboticrustacean (approved):
LGTM.

PR #626, opened 2026-05-10
fix(mobile): stop silently losing data on recipe and story edit (#615)

### LINKED ISSUE #615: Mobile: edit forms silently lose data on save (region + story image)
## Description
Two edit screens silently corrupt or drop data on save because state shape doesn't match the payload contract. Both are silent — no error, no toast, the user thinks the save worked.

### 1. RecipeEdit — region erased on every save
- **Files**: `app/mobile/src/screens/RecipeEditScreen.tsx:65`, `app/mobile/src/components/recipe/buildRecipeUpdateFormData.ts:18-24`
- The recipe detail API returns `region` as a friendly name string ("Aegean", "Black Sea"). The edit screen populates a TextInput with that string. On save, `buildRecipeUpdateFormData` does `Number(value)` → `NaN` → `Number.isFinite` is `false` → sends `region: null` to backend.
- **Impact**: Every recipe edit silently nulls the region. User changes a single typo and loses the region tag.

### 2. StoryEdit — re-uploads remote image URL as a local file
- **File**: `app/mobile/src/screens/StoryEditScreen.tsx:61, 135-136`
- When loaded for editing, `imageUri` is set to the remote URL. On save the URL is wrapped in FormData like a local file; on iOS this attempts a wasteful download-and-reupload, on Android it can fail outright.
- **Impact**: Saving any edit to a story that already has an image either corrupts the image, fails the save, or wastes bandwidth.

## Acceptance
- [ ] RecipeEdit either uses a region picker (preferred) or keeps the original region pk in state and only sends a new value when the user actually changes it
- [ ] StoryEdit only includes `image` in the FormData when the user has picked a NEW local file; otherwise omit it
- [ ] Manual test: edit a recipe without touching region → region survives
- [ ] Manual test: edit a story with an existing image, change only the title → image survives, no double upload

CaganIslam:
## Summary
Closes #615. Two edit screens were silently corrupting or dropping data on save because state shape didn't match the payload contract. Both bugs were invisible to the user — no error, no toast, the save "succeeded" but data was lost.

### 1. RecipeEdit silently nulled the region on every save
The recipe detail API exposes region as a friendly NAME string ("Aegean", "Black Sea"). The edit screen populated a TextInput with that string and on save did `Number(value)` → `NaN` → sent `region: null` to the backend. Every recipe edit nuked the region tag, even if the user only fixed a typo in the title.

**Fix**: drop the `region` field from the patch payload entirely. The TextInput is now read-only with a helper note ("Region cannot be changed from this screen yet. The original region is preserved."). Until a proper region picker lands, omitting the field from the body is the safest behaviour — the backend keeps the existing region.

### 2. StoryEdit re-uploaded the existing remote image as a local file on every save
When loaded for editing, `imageUri` was set to the remote URL. On save, that URL was wrapped in FormData like a local file — wasteful download-and-reupload on iOS, outright failure on some Android paths.

**Fix**: track `initialImageUrl` separately and only call `updateStoryImageById` when `imageUri !== initialImageUrl` (i.e., the user actually picked a new local file).

## Files
- `components/recipe/buildRecipeUpdateFormData.ts` — drop `region` parameter and field; comment explaining why.
- `screens/RecipeEditScreen.tsx` — region TextInput → `editable={false}` with helper text; drop `region` from the builder call.
- `screens/StoryEditScreen.tsx` — add `initialImageUrl` state, only upload image when it differs from the loaded value.

## Test
- `npx tsc --noEmit` clean
- Local docker stack:
  - Recipe edit on a `ayse@example.com` recipe with region "Black Sea": changed only the title → after save, region survived on the detail screen. Repeated 3 times — region stable.
  - Story edit on a story with an existing image: changed only the title → save fired exactly one PATCH (no second image upload). Then picked a new local image → save fired the JSON PATCH plus the image PATCH as expected.
- No regression on Create flows (untouched).

## Out of scope (follow-ups)
- A proper region picker on the recipe edit form so users can change region intentionally.
- Stories don't even have a region field on the create/edit form yet — separate issue coming.

Closes #615

### DIFF
--- a/app/mobile/src/components/recipe/buildRecipeUpdateFormData.ts
+++ b/app/mobile/src/components/recipe/buildRecipeUpdateFormData.ts
@@ -8,20 +8,21 @@ import type { LocalVideoSelection } from './RecipeVideoSection';
 export function buildRecipePatchJsonBody(input: {
   title: string;
   description: string;
-  region: string;
   qaEnabled: boolean;
   rows: AuthoringIngredientRow[];
 }): Record<string, unknown> {
   const validRows = input.rows.filter(
     (r) => r.ingredient.id != null && r.amount.trim() !== '' && r.unit.id != null,
   );
-  const regionTrim = input.region.trim();
-  const regionNum = regionTrim ? Number(regionTrim) : NaN;
 
+  // Region intentionally omitted from the patch body. The detail API exposes
+  // the region as a friendly NAME, not the FK pk we'd need to PATCH back.
+  // Including it here used to silently null the region on every edit because
+  // `Number("Aegean")` is `NaN`. Until a proper region picker lands, leaving
+  // the field out of the payload keeps the existing region untouched.
   return {
     title: input.title.trim(),
     description: input.description.trim(),
-    region: regionTrim && Number.isFinite(regionNum) ? regionNum : null,
     qa_enabled: input.qaEnabled,
     is_published: true,
     ingredients_write: validRows.map((r) => ({
--- a/app/mobile/src/screens/RecipeEditScreen.tsx
+++ b/app/mobile/src/screens/RecipeEditScreen.tsx
@@ -182,7 +182,6 @@ export default function RecipeEditScreen({ route, navigation }: Props) {
     const jsonBody = buildRecipePatchJsonBody({
       title,
       description,
-      region,
       qaEnabled,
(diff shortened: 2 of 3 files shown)


ErenCanOzkaya (approved):
LGTM

PR #629, opened 2026-05-10
feat(mobile): region picker on story create and edit forms (#627)

### LINKED ISSUE #627: Mobile: story create/edit forms don't expose the region field
## Description
The Story model has a direct optional `region` FK (added for map discovery in #381 — see `app/backend/apps/stories/models.py:47-48`), and the mobile StoryDetail screen renders the region as a pill (#561). But the **create and edit forms have no region field at all**, so authors can never tag a story with a region directly from mobile. The backend silently falls back to the linked recipe's region when present, which means stories without a linked recipe end up region-less and invisible on the map.

## Repro
1. Open Share → Create Story (or open one of your stories → Edit)
2. Look at the form — there's title, body, language, image, linked recipe, published toggle
3. **No region field anywhere**
4. Backend accepts `region: <pk>` on the serializer but mobile never sends it

## Wanted
- A region picker on both `StoryCreateScreen` and `StoryEditScreen` (similar shape to a select / single-choice modal — same UX as the future recipe region picker)
- Submit payload includes `region` when the user picks one
- StoryEdit pre-populates the picker from the loaded `story.region` (already surfaced as a friendly name; will need the pk too — surface `region_id` from `normalizeStoryDetail` for the picker)

## Files (likely)
- `app/mobile/src/screens/StoryCreateScreen.tsx`
- `app/mobile/src/screens/StoryEditScreen.tsx`
- `app/mobile/src/services/storyService.ts` — extend `normalizeStoryDetail` to surface `region_id` (or `regionPk`) for the picker
- New small component `RegionPicker` reusable on both story screens (and later recipe edit — see #615 follow-up)
- Backend already supports it; no backend work expected

## Acceptance
- [ ] Region picker appears on Create and Edit story forms
- [ ] Picker lists regions from `/api/regions/` (or `/api/map/regions/`)
- [ ] Edit pre-populates with the existing region
- [ ] Submit sends `region: <pk>` (or `null` when cleared)
- [ ] Story shows up on the map under the chosen region without needing a linked recipe
- [ ] No regression on stories that already have a linked-recipe-derived region

## Related
- #561 (region pill on Story detail — display side already done)
- #615 (recipe region read-only until proper picker; same picker should be reused there)
- Backend: `apps/stories/models.py:47-48` — region FK is optional, nullable, falls back to linked recipe's region

CaganIslam:
## Summary
Closes #627. The Story model has a direct optional `region` FK (added for map discovery in #381), and StoryDetail already renders the region as a pill (#561), but the create/edit forms had no region field at all. Authors couldn't tag a story with a region from mobile, which left stories without a linked recipe invisible on the map. This PR wires up a reusable region picker on both forms and surfaces the FK pk through the service layer so edit can pre-populate.

## Files
- `services/mapDataService.ts` — add `fetchRegions(): Promise<{id, name}[]>` helper for picker option lists. Uses the same `/api/regions/` endpoint already in use, just unwraps to a lighter shape.
- `services/storyService.ts` — `normalizeStoryDetail` now also surfaces `region_id` (the FK pk) alongside the friendly `region` label, so edit can drive the picker. `updateStoryById` payload accepts `region?: number | null`.
- `types/story.ts` — `region_id?: number | null` field.
- **New**: `components/pickers/RegionPicker.tsx` — modal-based picker. Trigger button shows the current selection (or "Pick a region…" placeholder); modal lists "None" + all regions with a check mark on the selected one; loading state; palette-safe (cream `bg` + `surfaceDark` border, `accentGreen` check mark). Reusable on the future recipe edit picker (see #615 follow-up).
- `screens/StoryCreateScreen.tsx` — `regionId` state, RegionPicker section above the linked recipe section, payload includes `region: regionId`.
- `screens/StoryEditScreen.tsx` — `regionId` + `regionLabel` (fallback) state, populated from `story.region_id` / `story.region` in `applyStory`; same picker section; payload includes `region: regionId`.

## Test
- `npx tsc --noEmit` clean
  - **Create**: Share → Story → tapped the region trigger → modal opened with the full region list (Black Sea, Aegean, etc.) → picked "Black Sea" → trigger updated → submit → detail showed the Black Sea pill.
  - **Edit**: opened an existing story (had no direct region, only the linked recipe's fallback) → region picker showed "Pick a region…" → picked one → save → detail now shows the chosen region pill.
  - **Re-edit**: opened that same edited story → region picker pre-populated with the saved value (proves `region_id` round-trips through `normalizeStoryDetail`).
  - **Clear**: picked "None" → save → detail falls back to the linked-recipe region (or shows no pill if neither set), as expected.
- No regression on stories that had a linked-recipe-derived region: still display the same way; only the explicit field is new.

## Out of scope
- A reusable picker on the recipe edit form (the `region` field on RecipeEditScreen is currently read-only per #615 — the same `RegionPicker` component should be wired up there in a follow-up).
- Web parity (#627 only mentions mobile in the title and acceptance).

Closes #627


ErenCanOzkaya (approved):
LGTM

Issue #511, opened 2026-05-09
Backend: Heritage Journey — HeritageJourneyStep model & API


CaganIslam:
**Parent:** #504

## Görev
- `HeritageJourneyStep` modeli oluştur:
  - `heritage_group` (FK → HeritageGroup)
  - `order` (PositiveIntegerField)
  - `location` (CharField)
  - `story` (TextField)
  - `era` (CharField, optional — "1700'ler", "Osmanlı dönemi" gibi)
- Serializer + ViewSet (CRUD)
- Admin panel kaydı
- Migration oluştur ve test yaz

PR #630, opened 2026-05-10
feat(backend): heritage group models, API and admin (#499)

laternis:
## Summary
- Add `apps/heritage/` with `HeritageGroup` and `HeritageGroupMembership` (generic FK to Recipe / Story).
- Read-only API at `/api/heritage-groups/` and `/api/heritage-groups/<id>/`.
- Recipe and Story serializers expose nested `heritage_group` (nullable).
- Django admin registers HeritageGroup with a generic inline for membership management.

## Test plan
- [x] `python manage.py test apps.heritage apps.recipes apps.stories` green
- [x] Manual: create a group in admin, attach a recipe and a story, hit `/api/heritage-groups/<id>/` and confirm both members appear
- [x] Recipe and Story API responses include `heritage_group` (null without membership, populated with one)

## Notes
- Foundation for #511 (HeritageJourneyStep) and #522 (CulturalFact). Those PRs will FK into HeritageGroup.

Closes #499.


roboticrustacean (approved):
merging now

PR #638, opened 2026-05-10
fix(mobile): tolerate empty bodies and malformed pagination links in httpClient (#618)

### LINKED ISSUE #618: Mobile: httpClient and pagination services crash on edge cases
## Description
Two robustness gaps in the HTTP layer cause crashes on perfectly valid backend responses.

### 1. apiPostJson / apiPatchJson crash on 204 No Content / empty body
- **File**: `app/mobile/src/services/httpClient.ts:55-66`
- Both functions unconditionally call `res.json()`. If the server returns 204, the body is empty and `JSON.parse('')` throws `SyntaxError`.
- **Impact**: Any endpoint returning an empty body crashes the calling screen. Affects mutations like delete, toggle, etc.

### 2. `new URL(data.next)` throws on malformed pagination link
- **Files**: `app/mobile/src/services/commentService.ts:33`, `app/mobile/src/services/tagsService.ts:18`
- No try/catch around `new URL(...)`. If the backend returns a relative or otherwise malformed `next` link, the constructor throws and the whole list fetch fails.
- **Impact**: Comments or tags can fail to load entirely on a single bad pagination link.

## Acceptance
- [ ] `apiPostJson`/`apiPatchJson`: check `res.status === 204` or `Content-Length === 0` before calling `.json()`; return `null` for empty bodies
- [ ] Pagination URL parsing wrapped in try/catch; on failure, treat as last page (don't blow up the list)
- [ ] Manual test: hit any 204-returning endpoint → no crash

CaganIslam:
## Summary
Closes #618. Two robustness gaps in the mobile HTTP layer would crash the calling screen on perfectly valid backend responses:

1. **Empty / 204 No Content body** — `apiGetJson` / `apiPostJson` / `apiPatchJson` unconditionally called `res.json()`. `JSON.parse('')` throws `SyntaxError: Unexpected end of JSON input`, so any 204-returning endpoint (some toggles, some delete-style mutations) would crash the screen.
2. **Malformed pagination `next` link** — `commentService.fetchCommentsForRecipe` and `tagsService.fetchAll` did `new URL(data.next)` with no try/catch. A single bad link took down the entire list fetch.

Both are edge cases — they may not be firing on the staging stack today, but the failure modes are obvious from the code and the cost of guarding them is tiny.

## Files
- `services/httpClient.ts` — add `parseJsonResponse<T>(res)` helper that short-circuits on 204 and on empty/whitespace body (returns `null as T`), and an exported `nextPagePath(next)` helper that wraps `new URL(...)` in try/catch and returns `null` (treated as last page) on parse failure. `apiGetJson`/`apiPostJson`/`apiPatchJson` now route through `parseJsonResponse`.
- `services/commentService.ts` — paginated `fetchCommentsForRecipe` uses `nextPagePath(data.next)` instead of constructing `new URL` inline.
- `services/tagsService.ts` — same fix in `fetchAll`.

## Test
- `npx tsc --noEmit` clean
- Verified the helpers in isolation with a small Node script (mimicking `Response` shapes):
  - `nextPagePath`: `null` / `undefined` / `''` / `'not a url at all'` / relative paths all return `null` (no throw); valid absolute URL returns its `pathname + search`.
  - `parseJsonResponse`: status 204 → `null`; empty body → `null`; whitespace-only body → `null`; valid JSON → parsed object.
  - Confirmed the before-fix behaviour: `new URL('not a url')` throws `Invalid URL`, `JSON.parse('')` throws `Unexpected end of JSON input`.
- Local docker stack regression check: opened a recipe → comments still load (the only paginating consumer of the helper); search filter chip rail → event/dietary tags still load; recipe edit save / comment post / vote toggle — no regression on the normal flows.

The edge cases themselves are hard to trigger from a healthy backend; the value is defensive — if/when the backend emits a 204 or a bad next link, the screen no longer crashes.

Closes #618

### DIFF
--- a/app/mobile/src/services/commentService.ts
+++ b/app/mobile/src/services/commentService.ts
@@ -1,4 +1,4 @@
-import { apiDelete, apiGetJson, apiPostJson } from './httpClient';
+import { apiDelete, apiGetJson, apiPostJson, nextPagePath } from './httpClient';
 
 export type CommentType = 'COMMENT' | 'QUESTION';
 
@@ -29,9 +29,7 @@ export async function fetchCommentsForRecipe(recipeId: string | number): Promise
       break;
     }
     collected.push(...data.results);
-    if (!data.next) break;
-    const url: URL = new URL(data.next);
-    path = `${url.pathname}${url.search}`;
+    path = nextPagePath(data.next);
   }
   return collected;
 }
--- a/app/mobile/src/services/httpClient.ts
+++ b/app/mobile/src/services/httpClient.ts
@@ -3,6 +3,36 @@ import { API_BASE_URL } from '../config/apiBase';
 
 const TOKEN_KEY = 'token';
 
+/**
+ * Read a JSON response safely. Returns `null` for 204 No Content and for any
+ * empty body — `res.json()` would otherwise throw `SyntaxError: Unexpected end
+ * of JSON input` on these inputs and crash the calling screen. Most mutation
+ * endpoints in this API do return JSON, so this only kicks in for the rare
+ * empty-body cases (delete-style responses, certain 200/204 toggles).
+ */
+async function parseJsonResponse<T>(res: Response): Promise<T> {
+  if (res.status === 204) return null as T;
+  const text = await res.text();
+  if (!text || !text.trim()) return null as T;
+  return JSON.parse(text) as T;
+}
+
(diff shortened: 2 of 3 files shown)


ErenCanOzkaya (approved):
LGTM

PR #639, opened 2026-05-10
fix(frontend): resolve all 9 critical Milestone 6 walkthrough bugs

### LINKED ISSUE #609: Web: Non-owners can access /recipes/:id/edit and /stories/:id/edit directly
## Summary

Both `/recipes/<id>/edit` and `/stories/<id>/edit` are not gated against non-owners at the route level. The pages mount, fetch the resource, render the edit form (briefly exposing the content), and only then display an inline authorization error. There is no early redirect.

## Steps to reproduce

1. Log in as user A.
2. Visit `/recipes/<id>/edit` or `/stories/<id>/edit` for a resource owned by user B.

## Expected

Immediate redirect to the detail page (or a clean 403) before any data is fetched or any form rendered.

## Actual

- `RecipeEditPage.jsx:184–186` — page fully loads, fetches recipe data, then shows an inline error after the fact.
- `StoryEditPage.jsx:107–109` — same pattern for stories.

## Related findings from the #540 walkthrough

- `StoryDetailPage` also renders a fake "Edit Story" button to non-owners that only sets a local error message on click (`StoryDetailPage.jsx:54–62`) — tracked separately.
- On the recipe Edit page, every save hard-codes `is_published: true` (`RecipeEditPage.jsx:156`), so a save by anyone forces publication.
- Zero-amount ingredients are silently dropped on submit (`RecipeCreatePage.jsx:133, 146`, `RecipeEditPage.jsx:135`) instead of being rejected with a validation error.
- On story edit, `linked_recipe: null` is never sent on PATCH (`StoryEditPage.jsx:83`), so unlinking a recipe is impossible.

Found during the #540 walkthrough.

### LINKED ISSUE #613: Web: /admin/moderation accessible to any logged-in user (no admin check)
## Summary

`ModerationPage` only checks `if (!user)` and otherwise renders the full moderation queue. There is no role/admin check. Any logged-in user can navigate to `/admin/moderation` and see the moderation UI. The backend may reject write actions, but the queue itself is exposed depending on the API.

## Steps to reproduce

1. Log in with a regular (non-admin) account.
2. Navigate directly to `/admin/moderation`.

## Expected

Non-admins are redirected (to `/`, `/login`, or a 403 page) before the queue is fetched.

## Actual

The page loads, calls `fetchModerationQueue`, and renders the full UI with Approve/Reject buttons.

## File references

- `app/frontend/src/pages/ModerationPage.jsx:28–33` (auth check is only `if (!user)`).
- `app/frontend/src/App.js:75–78` — the route uses `ProtectedRoute`, which only checks for a token.

## Related findings from the #540 walkthrough

- Approve/Reject buttons fire the API call immediately with no confirmation dialog (`ModerationPage.jsx:98–111`) — one misclick is enough to permanently approve or reject a tag.
- Failures from `approveTag` / `rejectTag` are silently swallowed: the `try { … } finally { … }` block has no `catch`, so the user sees the spinner clear with no error and the queue does not refresh (`ModerationPage.jsx:35–45`).

Found during the #540 walkthrough.

### LINKED ISSUE #614: Web: Onboarding "Finish" allows submitting with zero selections
## Summary

`OnboardingPage`'s `Finish` button is gated on `isComplete = STEPS.every((step) => Array.isArray(values[step.key]))`. Every step starts as `[]`, which is still an array, so the check is satisfied without any user input. A user can step through all four screens, select nothing, click Finish, and successfully POST empty arrays to `updateMe`.

## Steps to reproduce

1. Log in and navigate to `/onboarding`.
2. Click Next four times without selecting anything on any step.
3. Click Finish.

## Expected

`Finish` should remain disabled until at least one selection has been made (per step, or in total), with inline guidance for the user.

## Actual

`Finish` is enabled and submits empty arrays for `cultural_interests`, `regional_ties`, `religious_preferences`, and `event_interests` — defeating the personalization purpose of the flow.

## File references

- `app/frontend/src/pages/OnboardingPage.jsx:56–58` (`isComplete` check).
- `app/frontend/src/pages/OnboardingPage.jsx:155–162` (Finish button gating).

## Related findings from the #540 walkthrough

- "Skip for now" silently abandons onboarding without saving anything (`OnboardingPage.jsx:71–73`) — combined with the absent Profile page (tracked separately), there is no way for the user to come back and complete onboarding.
- The progress bar has only `aria-label="Onboarding progress"` and no `aria-valuenow` / `aria-valuemax`, so screen readers cannot announce progress (`OnboardingPage.jsx:97`).

Found during the #540 walkthrough.

ErenCanOzkaya:
## Summary

  Walkthrough-driven cleanup of every critical bug found during the Milestone 6 web walkthrough (#540). All 9 critical issues are addressed in this branch.

  Closes #606
  Closes #607
  Closes #608
  Closes #609
  Closes #610
  Closes #611
  Closes #612
  Closes #613
  Closes #614

  ## What changed

  ### Auth foundation (#608)
  - `AuthContext` now persists `refresh_token` to `localStorage` and exposes it via context.
  - `apiClient` has a response interceptor: 401 → POST `/api/auth/refresh/` (raw `axios`, not `apiClient`, to prevent self-await deadlock if the refresh itself 401s) → retry the
  original request. Parallel 401s share one in-flight refresh; refresh failure performs a hard logout + redirect to `/login`.

  ### Profile (#606)
  - Minimal `/profile` page: username + email view, `ContactabilityToggle` (moved from Inbox), Logout, "Edit cultural preferences" link to `/onboarding`. Throwaway pending Milestone 9
   `/users/:username` (#589).
  - Navbar `@username` is now a `Link` to `/profile`.

  ### Recipe / Story CRUD (#607, #609, #610)
  - `deleteRecipe` and `deleteStory` services + author-only Delete buttons with `window.confirm`, in-flight `Deleting…` disable to prevent double-clicks, and inline `role="alert"`
  error rendering so a failed delete no longer replaces the entire page.
  - `RecipeEditPage` and `StoryEditPage` redirect non-owners with `<Navigate replace />` before fetching, so the edit form never renders for the wrong user.
  - The fake non-author "Edit Story" button trap on `StoryDetailPage` is removed; non-authors now see no edit controls at all.

  ### Smaller correctness fixes
  - `MapPage` "See all from {region}" now passes `encodeURIComponent(region.name)` instead of `region.id` (#611).
  - `EventDetailPage` renders an "Event not found" panel with a back link instead of `return null` on 404 / fetch errors (#612).
  - `ModerationPage` redirects logged-in users without `is_staff` to `/` and short-circuits the queue fetch (#613).
  - `OnboardingPage` `isComplete` now requires at least one selection per step (`Array.isArray(v) && v.length > 0`) (#614).

  ### Misc
  - CI lint cleanup: `react-hooks/exhaustive-deps` on the new `useCallback` deps, `aria-selected` on a pre-existing `role="option"` button, and removal of an unused `toggleInList`
  helper.

  ## Tests

  - 293 / 293 passing across 37 suites (was 257 on `main`; +36 new tests).
  - New test files: `MapPage.test.jsx`, `EventDetailPage.test.jsx`, `ModerationPage.test.jsx`, `ProfilePage.test.jsx`.
  - Extended: `AuthContext`, `LoginPage`, `RegisterPage`, `api`, `recipeService`, `storyService`, `RecipeDetailPage`, `StoryDetailPage`, `RecipeEditPage`, `StoryEditPage`, `Navbar`,
  `OnboardingPage`.
  - `CI=true npm run build` succeeds.

  ## Test plan

  - [ ] Auth: log in → access token expires → next request silently refreshes; if refresh also fails, redirect to `/login`.
  - [ ] Profile: navbar `@username` opens `/profile`; ContactabilityToggle toggles; Log Out returns to `/login`.
  - [ ] Recipe & story delete: visible only to author; confirm cancel keeps content; confirm OK deletes + redirects; double-click is guarded; backend failure shows inline error
  without losing the page.
  - [ ] Direct nav to `/recipes/<other-user-id>/edit` or `/stories/<other-user-id>/edit` redirects to the detail page.
  - [ ] Map "See all from {Region}" lands on a populated search.
  - [ ] `/explore/99999999` shows "Event not found" with a back link.
  - [ ] `/admin/moderation` as non-staff redirects to `/`.
  - [ ] Onboarding Finish stays disabled until each step has ≥1 selection.

  ## Notes for review

  - The minimal `/profile` page is intentionally lightweight — Milestone 9 (#589 et al.) will replace it with a public `/users/:username` passport page.
  - 401-interceptor logic uses **raw `axios`** for the refresh POST specifically to avoid a self-await deadlock; the standalone `refreshAccessToken` helper in `authService.js` follows
   the same pattern for any future direct caller.


mustafaocakxyz (approved):
Lots of issues handled perfectly, good job

PR #636, opened 2026-05-10
feat(backend): cultural passport model, signal, and public endpoint (#583)

laternis:
## Summary
- New `apps/passport/` with `CulturalPassport` (OneToOne to User; level, total_points, active_theme).
- `post_save` signal creates a passport row for every new user; data migration backfills existing users.
- `GET /api/users/<username>/passport/` returns the full passport shape with stubbed sections (stamps, quests, timeline, culture_summaries) for #584 / #586 / #588 to fill later.

## Test plan
- [x] `python manage.py test apps.passport apps.users` green (74 tests)
- [x] `python manage.py test` full backend suite green (511 tests)
- [x] Manual: register a user, hit the passport endpoint, verify shape
- [x] Manual: anonymous request to a known user's passport returns 200

## Notes
- Built on top of #582 (public profile endpoint). PR is currently based on the `feat/backend/public-user-profile-582` branch and will be rebased onto `main` after #582 lands.
- `stamps`, `active_quests`, `timeline`, `culture_summaries` are currently empty arrays; #584, #586, #588 will populate them.
- `stats` counters (`recipes_tried`, `stories_saved`, `heritage_shared`, `cultures_count`) return 0 with a `TODO(#587)` until the tracking relations exist.
- `level_name` is derived from `level` via a small constant table in the serializer; keep in sync with #587 when real level logic lands.

Closes #583.

Issue #588, opened 2026-05-10
feat(backend): add passport timeline event log


laternis:
## Summary
Track significant passport events and surface them as a timeline in the passport API response.

## Model
`PassportEvent` (or a JSON field on the passport, choose what fits)
- `user` (FK)
- `event_type` (CharField, choices: recipe_tried / story_saved / stamp_earned / quest_completed / heritage_shared)
- `description` (CharField, human-readable)
- `timestamp` (DateTimeField)
- `related_recipe` (FK, nullable)
- `related_story` (FK, nullable)
- `stamp_rarity` (CharField, nullable)

## Endpoint
The `timeline` field of `GET /api/users/:username/passport/` includes the 50 most recent events, ordered newest first.

## Acceptance Criteria
- [ ] Each stamp trigger writes a corresponding event row
- [ ] Each quest completion writes an event
- [ ] Each level-up writes an event
- [ ] Heritage recipe submissions write an event
- [ ] Visitor mode includes the timeline by default (privacy toggle can land later)
- [ ] Tests cover the event creation triggers and the response trim to 50 events

## Notes
Open question per the spec: visitor visibility. Default to public; can be guarded later by a privacy setting on the user.

Issue #584, opened 2026-05-10
feat(backend): create Stamp model and passport action endpoints


laternis:
## Summary
Create a `Stamp` model and the two explicit passport action endpoints.

## Model
`Stamp`
- `user` (FK to User)
- `culture` (CharField, region/cuisine identifier)
- `category` (CharField, choices: recipe / story / heritage / exploration / community)
- `rarity` (CharField, choices: bronze / silver / gold / emerald / legendary)
- `earned_at` (DateTimeField)
- `source_recipe` (FK to Recipe, nullable)
- `source_story` (FK to Story, nullable)
- Unique constraint on (`user`, `culture`, `category`)

## Endpoints
- `POST /api/passport/recipes/:id/try/` — mark a recipe as tried, compute stamps, return updated passport summary
- `POST /api/passport/stories/:id/save/` — save a story to passport, compute stamps, return updated passport summary

## Stamp Upgrade Rules
| Rarity | Trigger |
|---|---|
| Bronze | First interaction with a culture |
| Silver | 3+ interactions with the same culture |
| Gold | 7+ interactions with at least 1 story saved |
| Emerald | User contributes their own content about the culture |
| Legendary | Heritage recipe with audio/video, 3-generation story, or rare combination |

When a new action is recorded, recompute the rarity for that (user, culture, category) and upgrade the existing stamp record rather than create a duplicate.

## Acceptance Criteria
- [ ] Migration creates the model with the unique constraint
- [ ] Both endpoints require authentication
- [ ] First call creates a Bronze stamp; subsequent calls upgrade to Silver / Gold based on counts
- [ ] Stamp upgrade is idempotent (re-trying the same recipe does not double-count)
- [ ] Response includes the updated passport summary so the client can refresh without a second request
- [ ] Trigger calls into points recompute and quest progress

## Notes
Depends on the passport model issue for the response shape. Triggers points recompute and quest progress logic in their respective issues.

Issue #582, opened 2026-05-10
feat(backend): add public user profile endpoint GET /api/users/:username/


laternis:
## Summary
Add a public endpoint that returns basic profile info for any user by username. Needed for the user profile page that hosts the Cultural Passport.

## Endpoint
`GET /api/users/:username/`

## Response Fields
- `username`
- `bio`
- `region`
- `cultural_interests`
- `religious_preferences`
- `event_interests`
- `created_at`
- `recipe_count` (computed)
- `story_count` (computed)

## Acceptance Criteria
- [ ] Endpoint resolves users by username (404 when not found)
- [ ] Response excludes sensitive fields: `email`, `is_contactable`, `password`
- [ ] Authenticated request for own username may include additional editable fields (TBD by author)
- [ ] Anonymous access returns 200 with public fields only
- [ ] Tests cover own-profile vs visitor cases and the 404 path

## Notes
Part of the Cultural Passport feature. Other backend pieces in this milestone: passport model and API, Stamp model and trigger endpoints, heritage flag, quest model, level/points, timeline.

PR #647, opened 2026-05-11
feat(backend): cultural passport model, signal, and public endpoint (#583)

laternis:
## Summary
- New `apps/passport/` with `CulturalPassport` (OneToOne to User; level, total_points, active_theme).
- `post_save` signal creates a passport row for every new user; data migration backfills existing users.
- `GET /api/users/<username>/passport/` returns the full passport shape with stubbed sections (stamps, quests, timeline, culture_summaries) for #584 / #586 / #588 to fill later.

## Test plan
- [x] `python manage.py test apps.passport apps.users` green (82 tests)
- [x] `python manage.py makemigrations --check` clean
- [x] Rebased onto current main; resolved `config/settings.py` and `config/urls.py` app-registration conflicts (heritage + passport both registered)

## Notes
- Supersedes PR #636, which merged into the `feat/backend/public-user-profile-582` branch rather than main. That branch is dead; this PR is the one that should land in main.
- Built on top of #582 (public profile endpoint), which is already in main.
- `stamps`, `active_quests`, `timeline`, `culture_summaries` are currently empty arrays; #584, #586, #588 will populate them.

Closes #583.


roboticrustacean (approved):
lgtm

PR #648, opened 2026-05-11
fix(mobile): use dedicated map endpoints for region pins and content (#620)

### LINKED ISSUE #620: Mobile: map and region content services use wrong/inefficient endpoints
## Description
Two map-related services hit the wrong endpoints and end up slow + incomplete. Both proper endpoints already exist on the backend (`/api/map/regions/` index, `/api/map/regions/<id>/content/` aggregated content), they're just not used.

### 1. mapDataService — wrong endpoint + N+1 count waterfall
- **File**: `app/mobile/src/services/mapDataService.ts:38`
- `fetchRegionPins()` hits `GET /api/regions/` (the lookup-table CRUD) instead of `GET /api/map/regions/` (the map endpoint with lat/lng + recipe counts). Coordinates are hardcoded in `regionGeo.ts` and each region's count requires a separate HTTP call.
- **Impact**: Map is slow to load on launch. Any region added on the backend that isn't in `regionGeo.ts` is silently invisible.

### 2. regionContentService — fetches all stories then filters client-side
- **File**: `app/mobile/src/services/regionContentService.ts:60-68`
- Fetches the entire stories list and filters by region locally instead of using `GET /api/map/regions/<pk>/content/`.
- **Impact**: Region detail sheet is slow and may show incomplete results (only page 1 again).

## Acceptance
- [ ] `fetchRegionPins` calls `/api/map/regions/` and uses backend lat/lng + count
- [ ] `regionGeo.ts` either deleted or kept only as a fallback for hardcoded display tweaks
- [ ] `regionContentService` calls `/api/map/regions/<pk>/content/`
- [ ] No client-side region filtering left in either service
- [ ] Manual test: every region pin shows on the map with the correct count; opening a region's content panel is fast

## Backend status
Verified during triage: `/api/map/regions/`, `/api/map/regions/<id>/`, `/api/map/regions/<id>/content/` all exist (`apps/map_discovery/urls.py`). No backend work needed.

CaganIslam:
## Summary
Closes #620. Two map-related services were hitting the wrong / inefficient endpoints:

1. **`mapDataService.fetchRegionPins`** called `GET /api/regions/` (the lookup CRUD) and then fired one extra `GET /api/recipes/?region=<name>` per region to count recipes — a classic N+1 waterfall. The proper map index endpoint already existed (`GET /api/map/regions/`) and returns everything in a single response, including content counts and seeded lat/lng.
2. **`regionContentService.fetchRegionContent`** fetched the *entire* `/api/stories/` list and filtered client-side by region — slow, page-1-only, and silently incomplete. The dedicated `GET /api/map/regions/<id>/content/` endpoint already supported `?type=recipe` / `?type=story` filtering on the server side.

Both endpoints already shipped on the backend (`apps/map_discovery/urls.py`); only the mobile client needed to switch over.

## Files
- `services/mapDataService.ts` — `fetchRegionPins` now calls `GET /api/map/regions/?geo_only=false` in a single request. Reads `content_count.recipes` directly, no per-region count call. Prefers backend `latitude`/`longitude` when populated, falls back to the hardcoded `regionGeo.COORDS` table otherwise (the seed data on staging hasn't populated lat/lng yet). The `countRecipesForRegion` helper is gone.
- `services/regionContentService.ts` — rewritten. `fetchRegionContent(regionId)` now hits `/api/map/regions/<id>/content/?type=recipe` and `?type=story` in parallel. Pulled in the response shape (`content_type`-tagged rows) from the new endpoint. Signature changed from `(regionName)` to `(regionId)` since the new endpoint keys on pk.
- `components/map/RegionDetailSheet.tsx` — accept a `regionId: number | null` prop alongside the existing `regionName: string | null` (still used for header / a11y labels). Fetch effect now depends on `regionId`.
- `screens/MapDiscoveryScreen.tsx` — pass `regionId={focused?.id ?? null}` through to the sheet.

## Test
- `npx tsc --noEmit` clean
  - Opened the Map tab → pins loaded in a single request (was: 1 region list + 1 count per region). Verified the recipe counts on each pin match what the backend reports under `content_count.recipes`.
  - Tapped a pin → RegionDetailSheet opened with the correct region's recipes and stories. Tab switching between Recipes / Stories is immediate. Tap-through to RecipeDetail / StoryDetail still works.
  - On staging the backend has no lat/lng seeded; the hardcoded `regionGeo.COORDS` fallback kicks in and pins still render in the right places. If/when the backend ships seeded coordinates, those automatically take over without a mobile change.

## Out of scope
- `regionGeo.COORDS` is intentionally kept as a fallback. Once the backend seeds lat/lng for all regions, that table can be removed in a follow-up — covered under M-8 polish (#622).
- The new `regionContentService` doesn't paginate yet; the dedicated endpoint already returns up to 20 items per type and the sheet doesn't surface a "load more" affordance. If it becomes a real limit, pagination is cheap to add.

Closes #620

### DIFF
--- a/app/mobile/src/components/map/RegionDetailSheet.tsx
+++ b/app/mobile/src/components/map/RegionDetailSheet.tsx
@@ -6,13 +6,15 @@ import { shadows, tokens, useTheme } from '../../theme';
 type Tab = 'recipes' | 'stories';
 
 type Props = {
-  /** Region name to load content for. Null/undefined hides the sheet. */
+  /** Region PK to load content for. Null/undefined hides the sheet. */
+  regionId: number | null;
+  /** Region display name for header/a11y labels. */
   regionName: string | null;
   onDismiss: () => void;
   onItemPress: (kind: Tab, id: string) => void;
 };
 
-export function RegionDetailSheet({ regionName, onDismiss, onItemPress }: Props) {
+export function RegionDetailSheet({ regionId, regionName, onDismiss, onItemPress }: Props) {
   const { accent } = useTheme();
   const [content, setContent] = useState<RegionContent | null>(null);
   const [loading, setLoading] = useState(false);
@@ -20,7 +22,7 @@ export function RegionDetailSheet({ regionName, onDismiss, onItemPress }: Props)
   const [tab, setTab] = useState<Tab>('recipes');
 
   useEffect(() => {
-    if (!regionName) {
+    if (regionId == null) {
       setContent(null);
       setError(null);
       return;
@@ -29,7 +31,7 @@ export function RegionDetailSheet({ regionName, onDismiss, onItemPress }: Props)
     setLoading(true);
     setError(null);
     setTab('recipes');
-    fetchRegionContent(regionName)
+    fetchRegionContent(regionId)
       .then((res) => {
         if (!cancelled) setContent(res);
       })
@@ -42,9 +44,9 @@ export function RegionDetailSheet({ regionName, onDismiss, onItemPress }: Props)
     return () => {
       cancelled = true;
     };
-  }, [regionName]);
+  }, [regionId]);
 
-  const visible = regionName != null;
+  const visible = regionId != null;
   const items = tab === 'recipes' ? (content?.recipes ?? []) : (content?.stories ?? []);
 
   return (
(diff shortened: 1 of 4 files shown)


mustafaocakxyz (approved):
Nice integration of backend, thank you

Issue #405, opened 2026-04-29
Mobile JWT: handle token_not_valid 401, add refresh + graceful logout


CaganIslam:
## Description                                                                
  `httpClient` attaches Bearer token from AsyncStorage but does not handle `401 
  { code: "token_not_valid" }`. Stale/expired/wrong-signature tokens cause      
  silent failures across all authenticated screens.                             
                                                            
  ## Scope                                                                      
  - Mobile (primary): 401 handling, refresh flow, AsyncStorage cleanup, redirect
   to Login.                                                                    
  - Backend (secondary): verify `/api/auth/token/refresh/` is wired; review
  `SIMPLE_JWT` lifetimes.                     
                                                                                
  ## Deliverables
  - On 401 `token_not_valid`: attempt refresh via `/api/auth/token/refresh/`    
  using stored refresh token, retry original request on success.
  - If refresh fails or absent: `AuthContext.logout()`, clear access + refresh  
  tokens, navigate to Login, toast "Session expired".       
  - Guard against retry loop on Login screen itself.                            
                                              
  ## Acceptance Criteria                                                        
  - [ ] Refresh succeeds → original request retried transparently, no UI flash.
  - [ ] Refresh fails → auto-logout + toast + Login redirect.                   
  - [ ] AsyncStorage cleared (both `access` and `refresh` keys).
  - [ ] Login screen 401 does not loop.                                         
  - [ ] Repro: login against backend A, point app at backend B with different   
  `SECRET_KEY` → auto-logout.                 
  - [ ] Repro: wait past `ACCESS_TOKEN_LIFETIME` (60 min) → refresh kicks in    
  silently.                                                                     
                                                                                
  ## References                                                                 
  - `app/mobile/src/services/httpClient.ts`                                     
  - `app/mobile/src/context/AuthContext.tsx`                
  - `app/backend/config/settings.py` (`SIMPLE_JWT`: ACCESS=60m, REFRESH=7d,     
  ROTATE=True)                                              
                                          
  Labels: area: frontend, area: backend, bug

PR #649, opened 2026-05-11
fix(mobile): close error feedback and recovery gaps in 4 mutations (#621)

CaganIslam:
## Summary
Closes #621. Four mutation paths had incomplete error/recovery flow — the UI optimistically updated, then either fell silent, left stale state, or leaked promise rejections. No single one was dramatic, but together they produced a "things sometimes go weird" feeling. All four are tightened here.

### 1. Comment delete cascaded replies in UI but not on the server
**File**: `components/recipe/RecipeCommentsSection.tsx:102`

The filter removed the target comment AND every reply (`c.parent_comment !== id`). Backend only deletes the single comment, so replies vanished from the UI and reappeared on the next reload. Filter now drops only the target; replies stay visible (orphaned) until either the user deletes them individually or backend cascade lands.

### 2. Comment vote failures were silent
**File**: `components/recipe/RecipeCommentsSection.tsx:122-132`

The catch block reverted the optimistic state but didn't tell the user. Their vote would flip and flip back without explanation. Added a toast on revert: "Could not save your vote. Please try again."

### 3. ContactabilityToggle had an `await` outside `try`
**File**: `components/ContactabilityToggle.tsx:21-37`

`await updateUser({ ...user, is_contactable: next })` was BEFORE the try block. `updateUser` calls AsyncStorage, which can throw on low-storage devices — that became an unhandled promise rejection that crashed the toggle. Moved the optimistic update inside try; the revert is now wrapped in its own try/catch so a second AsyncStorage failure doesn't blow up the recovery path either.

### 4. `logout()` never told the backend
**File**: `context/AuthContext.tsx` + `services/authService.ts`

Local-only logout. The refresh token was discarded at login (`data.refresh` ignored in `LoginScreen`), and there was no call to `POST /api/auth/logout/`. After "logout" the refresh token was still valid until natural expiry — a real security gap. Now:
- `AuthSuccess` type includes the optional `refresh` field
- `LoginScreen` and `RegisterScreen` pass `data.refresh` to `login()`
- `AuthContext.login()` accepts and stores `REFRESH_KEY` in AsyncStorage
- New helper `logoutRequest(refreshToken)` POSTs to `/api/auth/logout/` best-effort (catches errors, returns boolean)
- `AuthContext.logout()` retrieves the stored refresh, fires the request best-effort, then clears `TOKEN_KEY`, `REFRESH_KEY`, `USER_KEY`

If the server is unreachable or the refresh is already expired, local logout still happens — the user is never stranded.

## Test
- `npx tsc --noEmit` clean
  - **Delete cascade**: posted a reply to one of my comments, deleted the parent → reply remained visible (used to disappear and resurface on reload).
  - **Contactability**: toggled the switch repeatedly, including during a brief network blip — no crash, no orphaned save state, error surfaces in the existing inline banner.
  - **Logout**: signed out → backend log shows `POST /api/auth/logout/ 205`; signed back in cleanly. Attempted logout with backend down — local logout still completed, no crash, no toast spam.
  - **Vote toast**: verified by killing the backend mid-vote — toast appeared, vote flipped back, no console error.

## Out of scope
- Backend `POST /api/auth/logout/` already exists (`apps/users/views.py:LogoutView`). No backend work needed.
- A proper JWT silent-refresh interceptor on `apiPostJson` / `apiPatchJson` is still tracked under #405 — that's a separate concern from this PR.

Closes #621

### DIFF
--- a/app/mobile/src/components/ContactabilityToggle.tsx
+++ b/app/mobile/src/components/ContactabilityToggle.tsx
@@ -23,12 +23,19 @@ export function ContactabilityToggle() {
     setError(null);
     setSaving(true);
     const previous = enabled;
-    await updateUser({ ...user, is_contactable: next });
     try {
+      // Optimistic local update is INSIDE the try so an AsyncStorage failure
+      // (rare but real on low-storage devices) is caught instead of becoming
+      // an unhandled promise rejection that crashes the toggle.
+      await updateUser({ ...user, is_contactable: next });
       const updated = await updateMe({ is_contactable: next });
       await updateUser({ ...user, ...updated });
     } catch (e) {
-      await updateUser({ ...user, is_contactable: previous });
+      try {
+        await updateUser({ ...user, is_contactable: previous });
+      } catch {
+        // best-effort revert; the error toast below still surfaces the issue
+      }
       setError(e instanceof Error ? e.message : 'Could not update messaging preference.');
     } finally {
       setSaving(false);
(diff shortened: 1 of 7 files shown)


mustafaocakxyz (approved):
Nice guarding barriers, nice implementation

PR #650, opened 2026-05-11
fix(mobile): polish bundle — retry, cleanup, a11y, canonical field (#622)

CaganIslam:
## Summary
Closes #622 (partial). Bundle of small correctness, perf, and a11y fixes called out during the M6 mobile walkthrough audit. Each one is tiny on its own, but together they remove a long tail of "things sometimes go weird" rough edges before demo.

### Correctness
- **D — Search retry button was a noop** (`SearchScreen.tsx`): the retry handler was `setQuery((q) => q)`, which React bails out on. Added a `retryToken` state, included it in the search effect deps, and the Retry button now bumps it. Re-runs the search with the same query/filters.
- **Q — storyService sent the legacy `linked_recipe` field** (`storyService.ts`, `StoryCreateScreen.tsx`, `StoryEditScreen.tsx`): backend `StorySerializer.to_internal_value` has a "legacy" shim that maps `linked_recipe` → `linked_recipe_id`. Mobile now sends the canonical `linked_recipe_id` directly, so the day the shim is removed nothing breaks. Verified end-to-end with curl: POST `linked_recipe_id: 1` → backend stores it, GET returns `linked_recipe: 1` + recipe title.

### Cleanup / leaks
- **P — RecipeEditScreen `setTimeout` not cleaned up** (`RecipeEditScreen.tsx`): post-save `setTimeout(() => navigate('RecipeDetail'), 1500)` had no cleanup. If the user pressed back within 1.5s, they got bounced forward to RecipeDetail after the screen had already gone. Stored the timer id in a ref and clear it on unmount.
- **V — RecipeLinkPicker missing `cancelled` flag** (`components/story/RecipeLinkPicker.tsx`): the fetch effect had no unmount guard, so closing the picker mid-load triggered "Can't perform state update on unmounted component" warnings. Refactored the effect to use the standard `cancelled` pattern; the Retry action now bumps a `retryToken` state to re-fire the effect cleanly.
- **W — RecipeDetail unit-conversion wasteful re-render loop** (`RecipeDetailScreen.tsx`): `convertedByLine` was in the effect deps, so every batch result re-ran the effect (saw no missing items, exited). Drop it from deps with an explanatory comment — the toggle / recipe deps are sufficient triggers.

### A11y
- **Hit slop bump 6 → 10** across 7 files (pill buttons, cancel-link affordances, etc.). Apple HIG and Material both want ≥ 44pt minimum touch target; previous hitSlop of 6 still left some sub-target pills awkward to tap.

## Not done in this PR (intentional)

- **H — Onboarding "Finish" allows empty submit**: audit flagged this as a bug because `Array.isArray([])` is always true. After confirming with the product owner, the intent is leniency — users should be able to skip without selecting anything. Kept the current behaviour; added an inline comment explaining the choice so the next audit doesn't re-flag it.
- **N — Explore N+1 query waterfall**: re-reading `exploreService.fetchExploreCategories`, it already uses `Promise.all`. The audit was based on an older revision. No-op.
- **S — FlatList nested inside ScrollView** (`HomeScreen`, `ExploreScreen`): true perf issue but the refactor (single FlatList with section data + ListHeaderComponent) is too big for this bundle. Leaving for a separate PR so the rest of the polish can ship without holding for that change.

## Test
- `npx tsc --noEmit` clean
  - Searched with an active filter, killed the backend briefly, tapped Retry → search re-fired correctly (was a noop before).
  - Saved a recipe edit, immediately hit back → no unexpected RecipeDetail bounce.
  - Opened the story link picker, closed it during the fetch → no warning logged.
  - Toggled converted units on a recipe with many ingredients → no flicker on the toggle indicator; conversion happens once per line.
  - End-to-end backend test: `POST /api/stories/ {linked_recipe_id: 1, region: 1}` → 201, `GET /api/stories/<id>/` returns `linked_recipe: 1`, `recipe_title: "Black Sea Collard Green Sarma test"`, `region_name: "Aegean"`. Canonical field round-trips.
  - Onboarding: confirmed empty Finish still works (per product preference).

Closes #622

### DIFF
--- a/app/mobile/src/components/ContactabilityToggle.tsx
+++ b/app/mobile/src/components/ContactabilityToggle.tsx
@@ -57,7 +57,7 @@ export function ContactabilityToggle() {
         {saving ? 'Updating…' : enabled ? 'Allow new threads' : 'Block new threads'}
       </Text>
       {error ? (
-        <Pressable onPress={() => setError(null)} hitSlop={6} accessibilityRole="button">
+        <Pressable onPress={() => setError(null)} hitSlop={10} accessibilityRole="button">
           <Text style={styles.error}>{error} (tap to dismiss)</Text>
         </Pressable>
       ) : null}
--- a/app/mobile/src/components/home/StoryFeatureCard.tsx
+++ b/app/mobile/src/components/home/StoryFeatureCard.tsx
@@ -59,7 +59,7 @@ export function StoryFeatureCard({
               style={({ pressed }) => [styles.pill, pressed && styles.pressed]}
               accessibilityRole="link"
               accessibilityLabel={`Open profile of ${authorUsername}`}
-              hitSlop={6}
+              hitSlop={10}
             >
               <Text style={styles.pillText}>By {authorUsername}</Text>
             </Pressable>
@@ -70,7 +70,7 @@ export function StoryFeatureCard({
               style={({ pressed }) => [styles.recipePill, pressed && styles.pressed]}
               accessibilityRole="link"
               accessibilityLabel={`Open linked recipe ${recipeTitle}`}
-              hitSlop={6}
+              hitSlop={10}
             >
               <Text style={styles.recipePillText} numberOfLines={1}>
                 Recipe: {recipeTitle}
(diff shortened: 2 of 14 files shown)


mustafaocakxyz (approved):
Small but necessary changes for M6, great job!

PR #658, opened 2026-05-11
feat: add ?author=<id> filter to recipes and stories #637feat: add ?author=<id> filter to recipes and stories #637

### LINKED ISSUE #637: Backend: add ?author=<id> filter on /api/recipes/ and /api/stories/
## Description
The mobile UserProfile screen needs to show all recipes and stories authored by a specific user. Right now there is no way to ask the backend for that — `RecipeViewSet` and `StoryViewSet` don't expose a query-string filter on `author`, no `filter_backends` is configured, and there's no `/users/<id>/recipes/` or `/me/recipes/` route. The mobile screen has been working around this by fetching the global paginated list (page 1 only) and filtering by `author.id` client-side, which gives wrong results for any prolific user past page 1 — see #619.

## Repro
```bash
# both calls return the same 31 recipes — the ?author=1 filter is silently ignored
curl -H "Authorization: Bearer <token>" 'http://localhost/api/recipes/?author=1'
curl -H "Authorization: Bearer <token>" 'http://localhost/api/recipes/'
```

## Wanted
Add a server-side filter on the `author` FK pk for both endpoints:

- `GET /api/recipes/?author=<user_id>` — returns only recipes whose `author_id == <user_id>`
- `GET /api/stories/?author=<user_id>` — same for stories

Pagination should still work (DRF envelope `{count, next, previous, results}`). The filter should be combinable with existing query params (e.g. `?author=1&region=Aegean&page=2`).

## Implementation suggestion
Either:
- Add `django_filter_backends` + a `RecipeFilter` / `StoryFilter` class with `author = django_filters.NumberFilter(field_name='author__id')`, OR
- Override `get_queryset()` and read `self.request.query_params.get('author')` directly — simpler if no other filters needed.

## Acceptance
- [ ] `GET /api/recipes/?author=<id>` returns only that user's recipes
- [ ] `GET /api/stories/?author=<id>` returns only that user's stories
- [ ] Combining with existing filters (region, event, page) works
- [ ] Test added covering positive + negative case

## Mobile dependency
Closing #619 (mobile UserProfile shows wrong user / page-1 only) requires this filter. Mobile fix will land in a separate PR once this is shipped.

## Files
- `app/backend/apps/recipes/views.py` — `RecipeViewSet`
- `app/backend/apps/stories/views.py` — `StoryViewSet`
- Tests under each `tests.py`

roboticrustacean:
This PR adds server-side filtering by author ID for both the /api/recipes/ and /api/stories/ endpoints. This allows the mobile UserProfile screen to fetch content authored by a specific user efficiently, replacing the previous client-side filtering workaround.

- Updated apply_content_filters to handle author query parameter.
- Integrated apply_content_filters into StoryViewSet.
- Added automated tests for both models covering positive and negative cases.

Closes #637.


Emirhan777 (approved):
Filter to recipes and stories are applied correctly. I am approving this PR. LGTM.

PR #672, opened 2026-05-11
feat(frontend): Heritage Page, map, journey, Did You Know, calendar view (#500, #515, #517, #527, #668, #669)

Daglar1500:
## Summary

- **HeritagePage** (`/heritage/:id`): group name + description header, mixed recipe/story content grid, interactive heritage map (region-centric pins, polylines center→regions, click-to-panel showing that region's content), Heritage Journey timeline steps (#515), "Did You Know?" cultural fact cards (#517)
- **HeritageBadge** component: appears on `RecipeDetailPage` and `StoryDetailPage` when `heritage_group` is present in the API response, links to `/heritage/:id`
- **CalendarPage** (`/calendar`): monthly grid view, region + month filters, lunar event resolution with ☾ badge and italic subline, unresolved lunar events fall back to a Movable Feasts bucket (#669)
- **heritageService.js** + **calendarService.js**: service layers for `/api/heritage-groups/`, `/api/cultural-facts/`, `/api/cultural-events/`
- **AccountPage** (`/account`): view/edit the 4 onboarding preferences (cultural interests, regional ties, dietary preferences, event interests) with view/edit mode toggle; "My Account" added to navbar dropdown

## Test plan
- [ ] Navigate to `/heritage/:id` for a group that has members — verify header, content grid, map, journey steps, and Did You Know cards all render
- [ ] Click "Show Heritage Map" — verify center pin is the region with most recipes, polylines fan out to other regions, clicking a pin opens the side panel with that region's content
- [ ] Open a recipe detail page where `heritage_group` is set — verify HeritageBadge appears and links to correct Heritage Page
- [ ] Open a story detail page where `heritage_group` is set — same check
- [ ] Navigate to `/calendar` — verify events appear in correct month groups, lunar events show ☾ badge and italic subline
- [ ] Select a region filter — verify events filter correctly
- [ ] Click an event card — verify detail panel expands with description and linked recipes
- [ ] Open navbar dropdown — verify "My Account" appears between Profile and New Recipe
- [ ] Navigate to `/account` — verify saved preferences are pre-filled, edit mode shows chips, save persists

PR #673, opened 2026-05-11
feat(frontend): add /users/:username profile route with own vs visitor mode (#589)

Daglar1500:
## Summary

UserProfilePage (`/users/:username`): avatar initial, bio, region, join year, recipe/story count stats, cultural preference tags, and a passport placeholder shell ready for future issues (#591–#597).

Own profile mode (when `username === currentUser.username`) shows an "Edit preferences" link to `/account`. Visitor mode is fully read-only.

`passportService.js` created with `getPublicProfile` + stub functions for passport calls that will be wired up as backend issues #584–#588 land.

Author name links: navbar `@username` chip and author names on recipe/story detail pages now link to `/users/{username}`.

## Test plan
- [ ] Visit `/users/{username}` for an existing user — verify avatar, bio, region, stats, and preference tags render
- [ ] Visit `/users/{your_username}` while logged in — verify "Edit preferences" button appears
- [ ] Visit `/users/{other_username}` — verify no edit controls shown
- [ ] Visit `/users/doesnotexist` — verify 404 message and go-back button
- [ ] Click `@username` in navbar — verify navigates to own profile
- [ ] Open a recipe detail page — verify author name is a clickable link to `/users/{author_username}`
- [ ] Open a story detail page — same check

### DIFF
--- a/app/frontend/src/App.js
+++ b/app/frontend/src/App.js
@@ -19,6 +19,7 @@ import InboxPage from './pages/InboxPage';
 import ThreadPage from './pages/ThreadPage';
 import OnboardingPage from './pages/OnboardingPage';
 import ProfilePage from './pages/ProfilePage';
+import UserProfilePage from './pages/UserProfilePage';
 import MapPage from './pages/MapPage';
 import ExplorePage from './pages/ExplorePage';
 import EventDetailPage from './pages/EventDetailPage';
@@ -128,6 +129,7 @@ export default function App() {
             element={<ProtectedRoute><ModerationPage /></ProtectedRoute>}
           />
 
+          <Route path="/users/:username" element={<UserProfilePage />} />
           <Route path="/heritage/:id" element={<HeritagePage />} />
           <Route path="/heritage/:id/map" element={<HeritageMapPage />} />
           <Route path="/calendar" element={<CalendarPage />} />
--- a/app/frontend/src/__tests__/UserProfilePage.test.jsx
+++ b/app/frontend/src/__tests__/UserProfilePage.test.jsx
@@ -1,0 +1,91 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import { MemoryRouter, Route, Routes } from 'react-router-dom';
+import { AuthContext } from '../context/AuthContext';
+import UserProfilePage from '../pages/UserProfilePage';
+import * as passportService from '../services/passportService';
+
+jest.mock('../services/passportService');
+
+const mockProfile = {
+  username: 'alice',
+  bio: 'Loves Anatolian cuisine.',
+  region: 'Aegean',
+  created_at: '2024-03-01T00:00:00Z',
+  recipe_count: 4,
+  story_count: 2,
+  cultural_interests: ['Ottoman', 'Mediterranean'],
+  religious_preferences: ['Halal'],
+  event_interests: ['Ramadan'],
+};
+
+function renderPage(username, currentUser = null) {
+  return render(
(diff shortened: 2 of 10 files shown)


CaganIslam:
profile.username[0].toUpperCase() in UserProfilePage.jsx and Navbar.jsx will crash on empty string — use optional chaining (username?.[0]). Also <Link> nested inside <button> in Navbar is invalid HTML. No tests for the new UserProfilePage — please add at least basic coverage for own vs visitor mode.

ErenCanOzkaya (commented):
Thanks for putting this together — `UserProfilePage` itself is clean (404 path, own vs visitor mode, stats + tags + bio). A few items before this can land:

## Blocking — main rebase

This PR was forked off `feat/frontend/m8-heritage-calendar` (closed #672), so when it lands on `main` it **overrides the Milestone 8 web work that landed in #717**. Concretely the diff:

- **Deletes `HeritageMapPage.jsx` and its tests** → the dedicated `/heritage/:id/map` route disappears, replaced by an embedded region-centric map inside `HeritagePage`. The member-midpoint 🏛 marker + polylines from #500 are gone.
- **Renames the `HeritageBadge` prop** from `group` → `heritageGroup` → breaks `HeritageBadge.test.jsx`, `RecipeDetailPage.test.jsx`, `StoryDetailPage.test.jsx`.
- **Drops the `.recipe-heritage` and `.story-heritage` section wrappers** in favor of bare badges.
- **Removes the `CulturalFactCard` block on `RecipeDetailPage`** — `fetchCulturalFacts` + state + render are deleted, leaving the component orphaned. This regresses the AC on #517 ("Related fact display on recipe detail page").
- **Removes the Calendar navbar entry** → regresses #527.
- **Replaces `CalendarPage` with an older variant** that resolves lunar dates from a hard-coded yearly table client-side; the merged version delegates `?month=` filtering to the backend and routes lunar events to a dedicated panel.

Could you rebase against `main` and narrow this PR to just the `#589` scope?

```bash
git fetch origin main
git checkout feat/frontend/user-profile-route-589
git rebase origin/main      # take the main side on Heritage/Calendar conflicts
```

After the rebase the diff should be:

- `app/frontend/src/pages/UserProfilePage.{jsx,css}` (new)
- `app/frontend/src/services/passportService.js` (new)
- `app/frontend/src/App.js` (only the new `UserProfilePage` import + `/users/:username` route)
- `app/frontend/src/components/Navbar.jsx` (only the `@username` chip link)
- `app/frontend/src/pages/RecipeDetailPage.jsx` (only the `By <author>` → `<Link>` change)
- `app/frontend/src/pages/StoryDetailPage.jsx` (same)

`AccountPage` and `/account` already exist on `main`, so we don't need to touch them here.

## Non-blocking nits

- The `<span className="navbar-username">` is now a `<Link>` nested inside the menu `<button>`. `e.stopPropagation()` mitigates the dual click handler, but interactive content nested inside a button isn't valid HTML and trips a11y tools. Consider restructuring so the username chip sits as a sibling `<Link>` next to the button (which keeps avatar + chevron).
- The four passport stubs in `passportService.js` will lint-warn for unused args. Either rename them to `_username` or drop the param entirely until the real backend lands.
- `getPublicProfile` calls `extractApiError` for the 404 message but then overrides it with `'404'` and a hardcoded copy in the render. The current shape works, but it might be cleaner to return a typed result (`{ status: 'not_found' | 'error', message? }`) so callers don't compare error strings.

Once rebased + narrowed I'll re-review.

ErenCanOzkaya (approved):
LGTM

PR #676, opened 2026-05-11
fix(mobile): UserProfile uses server-side ?author filter and follows pagination (#619)

### LINKED ISSUE #619: Mobile: UserProfile shows wrong user content / only page 1 of recipes & stories
## Description
UserProfileScreen has two distinct bugs that compound to make profile pages unreliable.

### 1. useEffect missing `userId` dependency
- **File**: `app/mobile/src/screens/UserProfileScreen.tsx:41-67`
- The data-fetching effect only depends on `[reloadToken]`. When React Navigation reuses the same screen instance with different params (very common — tap user A's pill from feed, then user B's pill from a story), `route.params.userId` changes but the effect doesn't re-fire.
- **Impact**: Profile screen can show user A's recipes/stories with user B's header — wrong-user data on screen.

### 2. Fetches global recipe/story lists and filters client-side, page 1 only
- **File**: `app/mobile/src/screens/UserProfileScreen.tsx:47-76`
- Calls `fetchRecipesList()` and `fetchStoriesList()` (both unfiltered, single page) then filters by `author.id` on the client.
- **Impact**: Prolific users' profiles look empty or incomplete. Anyone whose content isn't on page 1 is invisible. As content grows, every profile gets worse.

## Wanted
- Fetch filtered server-side: `GET /api/recipes/?author=<id>` and `GET /api/stories/?author=<id>` (verify backend filter exists; if not, file a backend dependency)
- Effect deps: include `userId`
- Paginate fully or cap at a reasonable user-visible limit with a "Show all" affordance

## Acceptance
- [ ] Tap from user A profile to user B profile → screen shows B's content immediately, not A's
- [ ] User with 50+ recipes shows all of them on their profile (or paginates gracefully)
- [ ] No client-side `author.id` filtering left in the screen

## Backend dependency check
Need to confirm `?author=<id>` (or similar) filter is supported on `/api/recipes/` and `/api/stories/`. If not, file backend dependency before closing.

CaganIslam:
## Summary
Closes #619. UserProfileScreen had two distinct bugs that compounded into "wrong user content / silently incomplete":

1. **useEffect missing `userId` dependency** — navigating from profile A to profile B re-used the same screen instance with different params, but the data-fetching effect only depended on `[reloadToken]` so it never re-fired. The header showed B, but the recipes/stories still came from A.
2. **Global fetch + client-side filter, page 1 only** — `fetchRecipesList()` / `fetchStoriesList()` pulled the unfiltered first page and the screen filtered by `author.id` locally. Anyone whose content was on page 2+ was silently invisible.

The backend `?author=<id>` filter shipped in #637, so we can now ask the server for exactly the user's content and walk the DRF `next` link to gather all of it.

## Files
- `services/recipeService.ts` — `fetchRecipesList(filter?: { author? })` accepts the filter, sends `?author=<id>`, and walks `next` page-by-page (uses the existing `nextPagePath` helper from `httpClient`).
- `services/storyService.ts` — same pattern on `fetchStoriesList`.
- `screens/UserProfileScreen.tsx` — passes `userIdStr` to both services, adds `userIdStr` to the effect dep array, drops the now-redundant `authorIdOf` helper + `useMemo` client filter, and pulls `useMemo` out of imports.

## Test
- `npx tsc --noEmit` clean
- Verified counts match the database across 5 seeded users (per-API `?author=<id>` count exactly matches `Recipe.objects.filter(author=u).count()` and `Story.objects.filter(author=u).count()` for users 1-5: ayse 8/4, mehmet 3/2, elif 3/2, fatma 2/1, emily 0/0).
- On mobile, tapped from one user's profile to another via the recipe/story author pills — the screen immediately re-fetched and rendered the new user's content (used to keep showing the previous user's content until manual reload).
- Prolific author (ayse, 8 recipes): all 8 appear, not just page 1.

Closes #619

### DIFF
--- a/app/mobile/src/screens/UserProfileScreen.tsx
+++ b/app/mobile/src/screens/UserProfileScreen.tsx
@@ -1,5 +1,5 @@
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { Pressable, ScrollView, StyleSheet, Text, View } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import { ErrorView } from '../components/ui/ErrorView';
@@ -19,13 +19,6 @@ type ListItem = {
   author?: { id?: number | string; username?: string } | number | string | null;
 };
 
-function authorIdOf(item: ListItem): string | null {
-  const a = item.author;
-  if (!a) return null;
-  if (typeof a === 'object') return a.id != null ? String(a.id) : null;
-  return String(a);
-}
-
 export default function UserProfileScreen({ route, navigation }: Props) {
   const { userId, username } = route.params;
   const userIdStr = String(userId);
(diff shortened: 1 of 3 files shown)


Daglar1500 (approved):
i checked, well done

PR #679, opened 2026-05-11
feat(backend): allow profile field editing via PATCH /api/users/me/ (#659)

laternis:
## Summary
- PATCH /api/users/me/ now accepts username, bio, region, preferred_language alongside existing cultural-preference and is_contactable fields
- username uniqueness validated (400 on conflict, case-insensitive); email/role/is_staff/is_superuser stay blocked

## Test plan
- [x] python manage.py test apps.users

## Notes
- No migration: bio/region/preferred_language already on the User model
- Unblocks web #660 and mobile #661 profile editing screens

Closes #659.


roboticrustacean (approved):
lgtm

Issue #464, opened 2026-05-09
Zoom-to-region map view with per-recipe pins and unlocated-recipe bar (Web + Mobile + Backend)


CaganIslam:
## Description
Right now, tapping a region pin on `MapDiscoveryScreen` (mobile) or clicking a region on `MapPage` (web) opens the Search screen filtered by that region. Replace that flow with a richer in-map experience on **both platforms**, backed by dedicated API endpoints.

---

## Proposed Flow (Mobile + Web)

1. User taps/clicks a region pin on the discovery map.
2. The map **zooms in** so the entire region is visible (region bounds, not just a fixed delta).
3. The map shows **per-recipe pins** for every recipe in that region that has a location set.
4. Below the map (mobile) or in a sidebar panel (web), a **"Without a location"** section lists recipes from the same region that don't have coordinates, ordered alphabetically or by recency.
5. Tapping/clicking a recipe pin opens `RecipeDetail` for that recipe.
6. Tapping/clicking a card in the bottom bar / sidebar also opens `RecipeDetail`.
7. Back gesture / button returns to the full discovery map.

---

### New / Modified Fields
- `Recipe` model: add optional `latitude` / `longitude` fields (nullable `DecimalField`). Recipes without coordinates are valid — they appear in the "without location" section.
- `Region` model: add optional bounding box fields (`bbox_north`, `bbox_south`, `bbox_east`, `bbox_west` — nullable `DecimalField`) OR expose centroid + a sensible default delta.

### API Changes
- `GET /api/recipes/?region=<name>` — already works; ensure `latitude` and `longitude` are included in the serializer response when present.
- `GET /api/regions/<id>/` — include bounding box or centroid coordinates in the response.
- Consider a convenience endpoint: `GET /api/regions/<id>/recipes/` returning recipes split into `located` (with coords) and `unlocated` (without coords) groups.

### Migration
- Add migration for new nullable fields. No data loss — existing recipes get `null` for lat/lng.

---

### MapPage Enhancements (Leaflet / react-leaflet)
- On region click: `flyTo` the region bounds using `fitBounds` with the bounding box from the API.
- Render per-recipe `Marker` or `CircleMarker` for each located recipe in the region.
- Show a sidebar panel (or bottom drawer) listing unlocated recipes as clickable cards.
- Back button / "Show all regions" to return to the full discovery map view.
- Update `MapPage.css` for the new sidebar/drawer layout.

---

[...]

PR #683, opened 2026-05-11
feat(backend): zoom-to-region recipe coordinates, region bounds, region recipes endpoint (#662)

laternis:
## Summary
- Recipe gains latitude/longitude (nullable DecimalField); serialized on detail and ?region= list
- Region map bounds reuse the existing Region.bbox_* fields (Option A); no new Region field needed
- New GET /api/regions/<id>/recipes/ splits a region's recipes into located vs unlocated, newest first

## Test plan
- [x] python manage.py test apps.recipes
- [x] python manage.py test apps.recipes.tests.ZoomToRegionCoordsAPITest

## Notes
- Coordinate values are not seeded here; that is #657
- Migration filename carries the _662 suffix; if another recipes migration merges first this PR rebases and renumbers
- The new action overrides get_permissions on RegionViewSet so it stays publicly readable like list/retrieve (ModeratedLookupViewSet otherwise hardcodes IsAdminUser for non-CRUD actions)
- Backend foundation for #464 (web #468 + mobile)

Closes #662.

### DIFF
--- a/app/backend/apps/recipes/migrations/0022_zoom_region_coords_662.py
+++ b/app/backend/apps/recipes/migrations/0022_zoom_region_coords_662.py
@@ -1,0 +1,25 @@
+# Zoom-to-region: per-recipe map coordinates (#662).
+# Region map bounds reuse the existing Region.bbox_* fields (migration 0011),
+# so no Region change is needed here.
+
+from django.db import migrations, models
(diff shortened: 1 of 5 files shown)


roboticrustacean (approved):
looks good to me

PR #684, opened 2026-05-11
feat(backend): Beyond the Recipe RecipeCulturalContext model and nested API (#521)

laternis:
## Summary
- New RecipeCulturalContext model (OneToOne to Recipe) with seven narrative note fields
- Nested read/write on the Recipe serializer; admin inline registered

## Test plan
- [x] python manage.py test apps.recipes

## Notes
- No seed content here; that lands separately
- Migration filename carries the _521 suffix; rebases and renumbers if another recipes migration merges first
- Part of #509 (Beyond the Recipe), pairs with web #516 / mobile #525

Closes #521.


roboticrustacean (approved):
lgtm

Issue #20, opened 2026-02-12
Create my personal wikipage Ahmet Ayberk Durak


durakahmetayberk:
## Description
I am going to create / edit my personal page in the wiki.

## Motivation / Context
It is required in Lab #1

## Scope
Only one personal page in the wiki is in scope.

## Deliverables
One personal [page](https://github.com/bounswe/bounswe2026group12/wiki/Ahmet-Ayberk-Durak), done.

## Acceptance Criteria
- [x] Personal page contains enough information about the team member (me)
- [ ] Reviewer accepted the output

roboticrustacean:
Reviewed. Closing.

Issue #13, opened 2026-02-11
Create "About me" template for personal pages


mustafaocakxyz:
For the personal pages which will be presented in wiki, @roboticrustacean is assigned to create an "About me" template for every team member's use.

@mustafaocakxyz is the reviewer.

PR #712, opened 2026-05-11
fix(mobile): HomeScreen loading + focus refresh + fetchStoriesForRecipe pagination (#688)

CaganIslam:
## Summary
Closes #688. Three related mobile bugs around the home feed and recipe-linked stories.

### 1. HomeScreen had no loading state
- **File**: `app/mobile/src/screens/HomeScreen.tsx:22-57`
- Mount-time fetch had no `loading` flag, so the feed rendered as empty until the fetches resolved — users briefly saw "No stories yet" / "No recipes yet" on every cold open. Added an `initialLoading` boolean that gates a full-screen `LoadingView` until the first `loadFeed()` finishes.

### 2. HomeScreen never refreshed after mount
- Same file, `useEffect(...,[])` ran once on mount only. Publishing a new recipe / story on the Share tab didn't surface on return to Home; the user had to fully kill and reopen the app.
- Wrapped the load in a `useCallback` `loadFeed`, then hooked it to `useFocusEffect` so the feed re-fetches silently every time the Home tab regains focus (skipping the very first focus so the mount fetch doesn't double-fire). Also added a pull-to-refresh `RefreshControl` for the explicit case.

### 3. `fetchStoriesForRecipe` only read page 1
- **File**: `app/mobile/src/services/storyService.ts:93-97`
- Service called `GET /api/stories/` once and filtered by `linkedRecipeId` client-side. Stories whose linked recipe lived on page 2+ silently disappeared from the "Stories about this recipe" section.
- Now walks DRF `next` pages using the existing `nextPagePath` helper, then filters across the whole catalogue.

## Test
- `npx tsc --noEmit` clean
- **API contract verification**: walked `/api/stories/` page-by-page; total 18 stories in seed, 14 distinct linked recipes. Compared to old page-1 behaviour (page_size 5) which only sees 5 stories and 4 linked recipes — confirmed that recipes #20 and #13 each have 2 stories that the page-1 logic silently dropped. After the fix all 18 are walked and matched.
- HomeScreen UX changes verified on device:
  - Cold open shows the loading spinner instead of an empty-state flash.
  - Pull-to-refresh on Home reloads the feed.
  - Creating a new recipe on the Share tab → returning to Home → feed re-fetches and the new recipe is visible without a kill/relaunch.

## Files
- `app/mobile/src/screens/HomeScreen.tsx`
- `app/mobile/src/services/storyService.ts`

Closes #688

### DIFF
--- a/app/mobile/src/screens/HomeScreen.tsx
+++ b/app/mobile/src/screens/HomeScreen.tsx
@@ -1,7 +1,9 @@
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
-import { FlatList, Image, Pressable, ScrollView, StyleSheet, Text, TextInput, View } from 'react-native';
+import { useFocusEffect } from '@react-navigation/native';
+import { FlatList, Image, Pressable, RefreshControl, ScrollView, StyleSheet, Text, TextInput, View } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
-import { useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { LoadingView } from '../components/ui/LoadingView';
 import { shadows, tokens } from '../theme';
 import { fetchRecipesList } from '../services/recipeService';
 import { fetchStoriesList } from '../services/storyService';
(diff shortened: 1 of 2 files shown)


mustafaocakxyz (approved):
lgtm

PR #713, opened 2026-05-11
fix(mobile): auth-gate create screens and harden handleSkip (#691)

CaganIslam:
## Summary
Closes #691. Small form-UX bundle:
1. RecipeCreate and StoryCreate now render a sign-in gate when the user isn't authenticated, instead of letting them fill out the form and discover at submit-time that the API call fails.
2. `OnboardingScreen.handleSkip` wraps the AsyncStorage call in try/catch so a storage failure (rare on real devices, real on low-storage emulators) doesn't surface as an unhandled promise rejection — the user still exits the onboarding flow.

## Files
- `screens/RecipeCreateScreen.tsx` — import `useAuth`, render `authGate` view with heading + body + Log In / Register pills when `isReady && !isAuthenticated`.
- `screens/StoryCreateScreen.tsx` — same gate (StoryCreate already imported `useAuth` for the recipe picker; just destructured the missing fields).
- `screens/OnboardingScreen.tsx` — `handleSkip` wraps `AsyncStorage.setItem` in try/catch; falls through to `navigation.popToTop()` either way.
- `components/recipe/recipeFormStyles.ts` — shared `authGate` / `authGateHeading` / `authGateBody` styles so both create screens share the look.

## Test
- `npx tsc --noEmit` clean
- `handleSkip` simulated via Node test — even when `AsyncStorage.setItem` throws, `popToTop` still fires (no stuck-on-onboarding scenario).
- Code-path check: both create screens have `if (isReady && !isAuthenticated)` early return; the `isReady` guard prevents the bounce-before-auth-loads pattern that bit StoryEditPage on web (#694).
- Live test on device: signed out, opened both Share tabs → saw the sign-in gates instead of empty forms; tapped Log In / Register → routes correctly; signed back in → forms render normally.

Closes #691

### DIFF
--- a/app/mobile/src/components/recipe/recipeFormStyles.ts
+++ b/app/mobile/src/components/recipe/recipeFormStyles.ts
@@ -157,4 +157,22 @@ export const recipeFormStyles = StyleSheet.create({
     marginBottom: 8,
   },
   summaryLine: { fontSize: 15, color: tokens.colors.text, marginBottom: 4 },
+  authGate: {
+    flex: 1,
+    padding: 24,
+    justifyContent: 'center',
+    gap: 14,
+  },
+  authGateHeading: {
+    fontSize: 24,
+    fontWeight: '800',
+    color: tokens.colors.text,
+    fontFamily: tokens.typography.display.fontFamily,
+  },
+  authGateBody: {
+    fontSize: 15,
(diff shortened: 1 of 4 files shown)


mustafaocakxyz (approved):
lgtm

PR #714, opened 2026-05-11
fix(mobile): RecipeEdit validation parity with Create + show inline errors (#689)

CaganIslam:
## Summary
Closes #689. `RecipeEditScreen` was shipping weaker validation than `RecipeCreateScreen`, and two `attemptedSubmit={false}` props were hardcoded — so inline field errors never displayed on edit. Also, the patch-builder silently dropped unit-less ingredient rows.

## Bugs fixed
- **Validation gap**: edit only checked `title` (required) and `amount` (numeric). Create checks: description required, each ingredient row has an amount + ingredient + unit, and at least one row exists. Edit now mirrors all of those rules.
- **`attemptedSubmit={false}` hardcoded**: `<RecipeIngredientRowsSection>` and `<RecipeVideoSection>` were passed the literal `false`, so their inline errors never appeared on edit. Both now receive the live `attemptedSubmit` state.
- **Description input had no inline error block**: added the same `<InlineFieldError message={validation.description} />` pattern Create uses.
- **`buildRecipeUpdateFormData` filtered out unit-less rows**: backend accepts `unit: null`, so the filter was silently dropping otherwise-valid edits. Filter relaxed (still requires `ingredient` + `amount`); builder sends `unit: r.unit.id ?? null`.

## Files
- `screens/RecipeEditScreen.tsx`
- `components/recipe/buildRecipeUpdateFormData.ts`

## Test
- `npx tsc --noEmit` clean
- Validation logic verified via a Node simulation across 6 cases — all good, no description, no rows, row missing unit, negative amount, no title. Each one blocks/passes exactly as expected.
- Backend contract verified end-to-end: `PATCH /api/recipes/1/` with `ingredients_write: [{ingredient:1, amount:'1', unit:null}]` → **200 OK**. Backend accepts null unit; the old client filter was the silent-loss path.
- Live: edited a recipe with the description cleared → inline "Description is required." appears under the textarea (used to silently save). Cleared all ingredient rows → "Add at least one ingredient." top error. Added a row without a unit → "Choose a unit." row error. Filled everything → save works.

Closes #689

### DIFF
--- a/app/mobile/src/components/recipe/buildRecipeUpdateFormData.ts
+++ b/app/mobile/src/components/recipe/buildRecipeUpdateFormData.ts
@@ -11,8 +11,13 @@ export function buildRecipePatchJsonBody(input: {
   qaEnabled: boolean;
   rows: AuthoringIngredientRow[];
 }): Record<string, unknown> {
+  // Require an ingredient + a non-empty amount. The unit is *not* required —
+  // backend accepts `null` and we no longer silently drop unit-less rows
+  // (the previous filter did, swallowing valid edits). UI-level validation
+  // in RecipeEditScreen still asks for a unit; this filter only catches
+  // rows that slipped through (e.g. legacy data with a missing unit).
   const validRows = input.rows.filter(
-    (r) => r.ingredient.id != null && r.amount.trim() !== '' && r.unit.id != null,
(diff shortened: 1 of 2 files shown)


mustafaocakxyz (approved):
nice implementation

PR #715, opened 2026-05-11
fix(mobile): split create/edit into JSON + media phases, no orphan/duplicate on partial failure (#687)

### LINKED ISSUE #687: Mobile: recipe/story create partial failure leaves orphan + duplicate on retry
## Description
Recipe and Story create flows are two-step: (1) JSON POST creates the record, (2) optional multipart PATCH uploads media (image/video). If step 2 fails (network blip, big file, transient backend error), the record already exists on the backend but the screen shows a generic "Failed to publish. Please try again." Users naturally retry → step 1 fires again → **duplicate created**. The first half-uploaded record stays as an orphan with no media.

## Files
- `app/mobile/src/screens/RecipeCreateScreen.tsx:158-198`
- `app/mobile/src/screens/StoryCreateScreen.tsx:74-98`
- `app/mobile/src/screens/RecipeEditScreen.tsx:203-230` (analogous partial-failure on edit + media)

## Wanted
- Track which step failed. If step 1 throws → "Could not create, please retry." If step 1 succeeded and step 2 throws → message changes to "Recipe published but media upload failed — open it to retry the upload."
- After step 1 succeeds, **navigate to the detail screen** regardless of step 2 outcome. The user lands on the new item and can re-upload media there; no second create is even possible from that screen.
- For RecipeEdit: same idea — JSON PATCH succeeded → don't retry the JSON when only media failed.

## Acceptance
- [ ] No duplicate recipe/story creation when media upload fails on retry
- [ ] Toast distinguishes "create failed" vs "media upload failed"
- [ ] After successful JSON create, user navigates to the detail (current Recipe behaviour is already correct after #616; Story create needs parity)
- [ ] Edit: only re-PATCH the failing piece (JSON or media), not both

CaganIslam:
## Summary
Closes #687. Recipe and Story create flows are two-step: (1) JSON POST creates the record, (2) optional multipart PATCH uploads media. If step 2 failed, the previous code showed a generic "Failed to publish" toast even though the record already existed on the backend — users naturally retried, fired another POST, and ended up with a duplicate + an orphan record. RecipeEdit had the analogue across three PATCHes (JSON + image + video).

## Fix
Each step is now its own `try`, and a failed later step does **not** retry or roll back a successful earlier step.

### RecipeCreate / StoryCreate
- Step 1 fails → toast says "Could not publish", screen stays on the form (no orphan, retry-safe).
- Step 1 succeeds → `createdId` is captured.
- Step 2 (media) fails → soft fail: toast says "published — but media upload failed. Open it to retry from edit."
- **Always navigate to the detail screen after step 1 succeeds.** The user lands on the new record; the only way to "retry" is from the edit screen, so duplicate POST is impossible.

### RecipeEdit
- JSON PATCH fail → "Could not save", stay on form, no further attempts.
- JSON PATCH succeeds → image and video PATCHes attempted independently.
- Successful media PATCHes clear `localImage` / `localVideo` so a retry doesn't re-upload them.
- Toast names which piece didn't save: "image upload failed" / "video upload failed" / "image and video failed".
- Navigate to detail after the JSON succeeds, regardless of media outcome.

## Files
- `screens/RecipeCreateScreen.tsx`
- `screens/StoryCreateScreen.tsx`
- `screens/RecipeEditScreen.tsx`

## Test
- `npx tsc --noEmit` clean
- Logic simulated in Node across 4 cases (happy with media, happy no media, POST fails, POST ok / media fails) — toasts and navigation behave exactly as expected.
- Backend contract verified end-to-end: `POST /api/recipes/` returns 201 with id; subsequent bad multipart `PATCH /api/recipes/<id>/` returns 400; **the recipe persists** (`GET .../` 200). Confirms the orphan scenario was real and the fix's "always navigate after step 1" approach correctly takes the user to the existing record instead of leaving them to spawn a duplicate.

## Mobile parity
Web #703 covers the analogue partial-failure on `RecipeCreatePage`; same pattern applies there.

Closes #687

### DIFF
--- a/app/mobile/src/screens/RecipeCreateScreen.tsx
+++ b/app/mobile/src/screens/RecipeCreateScreen.tsx
@@ -156,8 +156,9 @@ export default function RecipeCreateScreen({ navigation }: Props) {
     };
 
     void (async () => {
+      // Step 1 — JSON create. Failure means nothing exists, retry is safe.
+      let createdId: number | null = null;
       try {
-        // Same as web: JSON create, then multipart PATCH for video/thumbnail.
         const created = await apiPostJson<{ id: number }>('/api/recipes/', {
           title: title.trim(),
           description: payload.description,
@@ -169,10 +170,22 @@ export default function RecipeCreateScreen({ navigation }: Props) {
             unit: x.unit?.id ?? x.unit,
           })),
         });
-        if (payload.video || payload.image) {
+        createdId = created.id;
+      } catch {
+        showToast('Could not publish recipe. Please try again.', 'error');
+        setSubmitting(false);
+        return;
+      }
+
+      // Step 2 — multipart media upload. A failure here MUST NOT trigger a
+      // retry of step 1 (that creates an orphan + duplicate). We treat it as
+      // a soft failure: the recipe exists, user lands on its detail, and a
+      // toast tells them to re-upload from edit.
+      let mediaFailed = false;
+      if (payload.video || payload.image) {
+        try {
           const fd = new FormData();
           if (payload.video) {
-            // React Native file upload shape (not a web Blob).
             fd.append('video', {
               uri: payload.video.uri,
               name: payload.video.fileName ?? 'recipe-video.mp4',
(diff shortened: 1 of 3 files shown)


mustafaocakxyz (approved):
good job

PR #719, opened 2026-05-11
feat(frontend): floating chat tray — X.com-style docked tab with inline messaging (#716)

Daglar1500:
## Summary

- **Chat tray**: Fixed to the bottom-left corner, always visible as a compact tab bar (💬 Messages + unread badge + chevron). Only shown when logged in.
- **Thread list**: Clicking the tab opens a panel upward showing recent conversations sorted by last message. Each row shows avatar, username, last message preview, relative time, unread count, and a green dot if the last message was within 5 minutes.
- **Inline mini chat**: Clicking a thread opens a mini chat window directly in the tray — no navigation to `/inbox`. Shows message history, send input (Enter to send / Shift+Enter for newline), and a back button to return to the thread list.
- **ChatContext**: Fetches threads on mount and polls every 30 seconds for new messages.
- **Navbar**: Removed "Inbox" from the dropdown — the chat tray replaces it.

## Test plan
- [ ] Login → compact "💬 Messages" tab visible at bottom-left
- [ ] Logout → tab disappears
- [ ] Click tab → thread list opens upward; click again → closes
- [ ] Unread messages → red badge on tab and on the thread row
- [ ] Click a thread → mini chat opens inside the tray, messages load
- [ ] Send a message → appears in the chat immediately
- [ ] Back button → returns to thread list
- [ ] "View all messages →" → navigates to `/inbox`
- [ ] Click outside the panel → closes
- [ ] Navbar dropdown no longer shows "Inbox"

### DIFF
--- a/app/frontend/.env.test
+++ b/app/frontend/.env.test
@@ -1,0 +1,2 @@
+REACT_APP_API_URL=http://localhost:8000
+REACT_APP_USE_MOCK=false
--- a/app/frontend/Dockerfile
+++ b/app/frontend/Dockerfile
@@ -11,6 +11,8 @@ COPY . .
 
 ARG REACT_APP_API_URL=http://localhost
 ENV REACT_APP_API_URL=$REACT_APP_API_URL
+ARG REACT_APP_USE_MOCK=false
+ENV REACT_APP_USE_MOCK=$REACT_APP_USE_MOCK
 RUN npm run build
 
 # ---- Stage 2: serve ----
--- a/app/frontend/src/App.js
+++ b/app/frontend/src/App.js
@@ -1,6 +1,7 @@
 import { Routes, Route, useLocation } from 'react-router-dom';
 import Navbar from './components/Navbar';
 import SkipLink from './components/SkipLink';
+import ChatTray from './components/ChatTray';
 import ProtectedRoute from './components/ProtectedRoute';
 import GridMotion from './components/GridMotion';
 import HomePage from './pages/HomePage';
@@ -71,6 +72,7 @@ export default function App() {
         </>
       )}
       <Navbar />
+      <ChatTray />
       <div className="page-wrapper" id="main-content">
         <Routes>
           <Route path="/" element={<HomePage />} />
--- a/app/frontend/src/__tests__/ChatContext.test.jsx
+++ b/app/frontend/src/__tests__/ChatContext.test.jsx
@@ -1,0 +1,89 @@
+import { act, render, screen, waitFor } from '@testing-library/react';
+import { useContext } from 'react';
+import { ChatContext, ChatProvider } from '../context/ChatContext';
+import { AuthContext } from '../context/AuthContext';
+import * as messageService from '../services/messageService';
+
+jest.mock('../services/messageService');
+jest.useFakeTimers();
+
+const mockThreads = [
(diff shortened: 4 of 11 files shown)


ErenCanOzkaya (commented):
Nice work on the docked tray + inline mini chat — the X.com-style toggle, polling cadence, and "recently active" green dot all land well, and the AC list is covered. A few things to address before merge:

## Blocking — main rebase (silent regression)

This PR is `MERGEABLE` per the GitHub check, but the diff against `main` deletes lines that exist on the merged Milestone 8 work:

- **`App.js`**: removes `import AccountPage` + the `/account` `<Route>` → `/account` URL falls through to `NotFoundPage`.
- **`Navbar.jsx`**: removes the Calendar nav entry (regresses #527 AC) and the "My Account" dropdown link.

Looks like this branch was forked off `feat/frontend/m8-heritage-calendar`, so it carries the deletes from there. Removing the "Inbox" dropdown entry is fine — that's a deliberate part of #716 since the tray's "View all messages →" replaces it. Calendar / My Account / `/account` should stay.

```bash
git fetch origin main
git checkout feat/frontend/chat-tray-716
git rebase origin/main      # keep main's side on App.js + Navbar.jsx conflicts
```

Then reapply *only* the Inbox dropdown removal as a separate hunk on top of `Navbar.jsx`.

## Non-blocking — tests

`ChatTray.jsx` (~235 lines), `ChatContext.jsx`, and the helper utilities ship without any Jest coverage. M8 set a precedent of test-per-component / per-service in this part of the repo. At minimum:

- `ChatContext` provider — threads init on mount, polling kicks in after 30s, `markRead` zeroes `unreadCount` locally and calls the service.
- `ChatTray` — renders nothing when `user` is null, renders the tab with unread badge when threads have unread, opens/closes panel on toggle, click-outside closes it.
- Pure helpers `relativeTime` and `isRecentlyActive` — easy unit tests, deterministic with a frozen `Date.now()`.

`.env.test` correctly forces `REACT_APP_USE_MOCK=false` so tests would target the real services; the existing service mocks in other test files give you the pattern to copy.

## Non-blocking — small notes

- `ChatContext.jsx` exposes `markRead` but the conversation view also calls `markThreadRead` directly inside its own `useEffect`. The context already does the optimistic local update on `handleThreadClick`; double-calling `markThreadRead` on conversation open is a redundant network hit. Could route everything through the context.
- `handleKeyDown` in `ChatConversation` carries `// eslint-disable-next-line react-hooks/exhaustive-deps` because `handleSend` isn't memoized. Wrapping `handleSend` in `useCallback` (or inlining it) would let you drop the disable.
- `relativeTime` returns `${diff}s` even for negative values when the system clock skews backwards or when the API returns timestamps slightly in the future. Clamp `diff = Math.max(0, ...)` so you never render `-3s`.
- `setInterval` cleanup looks right (`clearInterval(intervalRef.current)` runs on user change), but the new interval is scheduled inside an effect that re-runs whenever `refresh` changes identity. With `useCallback([user])` `refresh` is stable per user, so this is fine in practice — worth a one-liner test to lock it down.

Once rebased + tests added I'll re-review.

ErenCanOzkaya (approved):
Backend tests failed on github actions due to timeout. Works fine in local, LGTM

PR #725, opened 2026-05-11
feat(mobile): add zoom in/out controls to map screens (#722)

CaganIslam:
Closes #722.

## Summary

Pinch-zoom on `react-native-maps` works but isn't discoverable. This PR adds an explicit floating `＋` / `－` control overlaid on every map view in the app.

## What changed

- New component `src/components/map/MapZoomControls.tsx`
  - Accepts a `mapRef: React.RefObject<MapView | null>` prop.
  - Two stacked `Pressable`s in a single rounded card (cream `bg`, `surfaceDark` border, `md` shadow). Positioned absolutely top-right by default; consumer can override via `style`.
  - On press: calls `mapRef.current.getCamera()`. On Android, `Camera.zoom` is populated, so it animates the camera by `±1` clamped to `[3, 18]`. On iOS (no `zoom` on the camera object), falls back to `getMapBoundaries()` + `animateToRegion(...)` with halved/doubled `latitudeDelta` / `longitudeDelta`, clamped to a sane envelope.
  - Accessibility: `accessibilityRole="button"`, labels `"Zoom in"` / `"Zoom out"`, `hitSlop={10}`.
- Wired into the two map screens that exist today:
  - `MapDiscoveryScreen` — added a `mapRef`, attached it to the `MapView`, and rendered `<MapZoomControls>` inside the fill view.
  - `HeritageMapScreen` — already had a `mapRef`; placed the controls at `top: 90` so they sit directly below the existing legend card.

`RegionMapDetailScreen` from the issue description doesn't exist in this codebase, so there was nothing to wire there.

## Notes

- No new dependencies.
- Pinch / double-tap zoom is untouched (`zoomEnabled` / `scrollEnabled` not set).
- Only stable palette tokens used: `bg`, `surfaceDark`, `text` (no `primary*` per #617).
- `npx tsc --noEmit` passes.

## Test plan

- [ ] Open the Map tab → controls appear top-right of the map.
- [ ] Tap `＋` → camera animates one zoom step in.
- [ ] Tap `－` → camera animates one zoom step out.
- [ ] Spam-tap `＋` past street level → no crash, motion stops at the max clamp.
- [ ] Spam-tap `－` past world view → no crash, motion stops at the min clamp.
- [ ] Pinch-zoom still works normally alongside the buttons.
- [ ] Open a Heritage group map → controls appear under the legend card.
- [ ] VoiceOver / TalkBack: each button announces "Zoom in" / "Zoom out".

PR #723, opened 2026-05-11
feat(mobile): zoom into region with per-recipe pins and unlocated bar (#464)

CaganIslam:
## Summary
Closes #464 (mobile portion). Tapping a region pin on `MapDiscoveryScreen` used to open a small sheet (regions overview). Now it pushes a dedicated `RegionMapDetailScreen` that **zooms in on that region** and plots **one marker per recipe with coordinates**, with a horizontal "Without a location" bar at the bottom for recipes from the same region that don't have lat/lng set.

## Files
- **New** `screens/RegionMapDetailScreen.tsx` — `MapView` initialised from the region's bbox (centroid + delta fallback when bbox is missing); `fitToCoordinates` re-fits to the actual recipe pins after data loads. Per-recipe `Marker` with callout (title + author + "Tap to open →"). Floating header card with the region name and counts. Horizontal `FlatList` of unlocated recipes pinned to the bottom; each card has thumb + title + author and routes to RecipeDetail. Loading / error / empty states handled.
- `services/mapDataService.ts` — new `fetchRegionRecipes(regionName)` that:
  - Reads bbox + centroid from `/api/map/regions/`
  - Walks `/api/recipes/?region=<name>` via the `next` cursor (handles pagination)
  - Splits into `located` (recipes with lat/lng) and `unlocated` (without)
  - **Defensive number parsing**: backend serialises `Recipe.latitude/longitude` as DRF DecimalField strings ("38.500000") while Region bbox comes through as floats. The service `parseFloat`s both so the screen plots either shape correctly.
- `navigation/types.ts` + `PublicStackNavigator.tsx` — register `RegionMapDetail: { regionId, regionName }`.
- `screens/MapDiscoveryScreen.tsx` — pin tap now also `navigation.navigate('RegionMapDetail', …)` alongside the existing region focus state (the old `RegionDetailSheet` stays mounted but isn't the primary drill-in anymore; can be removed in a follow-up once stories get their own surface).

## Test
- `npx tsc --noEmit` clean
- API contract verified: `/api/recipes/?region=Aegean` returns 3 recipes (2 with lat/lng, 1 without). `/api/map/regions/1/` returns bbox `{north: 39.5, south: 37.0, east: 28.5, west: 26.0}` and centroid `(38.4192, 27.1287)`.
[...]

PR #733, opened 2026-05-11
feat(mobile): wire MapZoomControls into RegionMapDetailScreen (follow-up to #722 / #464)

CaganIslam:
## Summary
Small follow-up promised in PR #725. `MapZoomControls` (#722) landed but only got wired into `MapDiscoveryScreen` and `HeritageMapScreen` — `RegionMapDetailScreen` was added in #723 (PR #723) which wasn't on `main` at the time. Now that both are on `main`, this PR adds the same controls to RegionMapDetailScreen.

## File
- `screens/RegionMapDetailScreen.tsx` — import `MapZoomControls`, render `<MapZoomControls mapRef={mapRef} style={styles.zoomControls} />` between the MapView and the header overlay. The `zoomControls` style positions it at `top: 90` so it sits below the existing region-name header card (which lives at `top: 16`).

## Test
- `npx tsc --noEmit` clean.
- Visual: open a region pin from MapDiscovery → RegionMapDetailScreen → zoom in/out buttons appear below the floating header card; +/- animate the camera one zoom step at a time, with the same `[3, 18]` clamps as the other screens.

Closes (follow-up to) #722 and #464.

### DIFF
--- a/app/mobile/src/screens/RegionMapDetailScreen.tsx
+++ b/app/mobile/src/screens/RegionMapDetailScreen.tsx
@@ -5,6 +5,7 @@ import MapView, { Marker } from 'react-native-maps';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import { ErrorView } from '../components/ui/ErrorView';
 import { LoadingView } from '../components/ui/LoadingView';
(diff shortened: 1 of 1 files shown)


mustafaocakxyz (approved):
great job

PR #745, opened 2026-05-11
feat(backend): seed cultural facts catalogue for Did You Know cards (#664)

laternis:
## Summary
- Add curated region-tied CulturalFact data (~93 facts, 3 per region across the 31 broad cuisine regions that seeded recipes use)
- Add a standalone, idempotent seed_cultural_facts command that resolves each entry's region by name, leaves heritage_group null, and warns and skips unknown regions instead of crashing
- Add apps/heritage/tests_facts_seed.py covering count, per-region coverage, heritage_group-null invariant, idempotency, unknown-region skipping, ?region= filter, and /random/

## Test plan
- [x] cd app/backend && python manage.py seed_canonical && python manage.py seed_cultural_facts
- [x] python manage.py test apps.heritage.tests_facts_seed
- [x] python manage.py test apps.heritage (67 tests, all green)
- [x] python manage.py makemigrations --check --dry-run (clean)

## Notes
- Region-tied facts only (heritage_group left null); the four heritage groups already had nested facts via seed_canonical, so region coverage was the gap. Standalone fixtures/cultural_facts.json and command to avoid contention on seed_canonical.json.
- "Caucasian" appears in the fixture but is not currently a seeded region, so the command warns and skips it; this also exercises the missing-region path on every run.

Closes #664.

### DIFF
--- a/app/backend/apps/heritage/management/commands/seed_cultural_facts.py
+++ b/app/backend/apps/heritage/management/commands/seed_cultural_facts.py
@@ -1,0 +1,92 @@
+"""Idempotent seeder for region-tied "Did You Know?" cultural facts (#664).
+
+The CulturalFact model already ships with a handful of facts nested under
+the four heritage groups (seeded via seed_canonical). What was missing is
+facts tied to a Region, which the mobile "Did You Know?" cards and the web
+parity work query per recipe region. Without them the section silently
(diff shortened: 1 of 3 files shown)


roboticrustacean (approved):
thanks

PR #746, opened 2026-05-11
test(backend): API contract verification suite (#535)

laternis:
## Summary
- Add `app/backend/apps/common/tests_api_contract.py` asserting the response contract for every endpoint group named in #535: auth, recipes, stories, search, map, messaging, notifications, cultural content, heritage, lookups
- Assertions cover status codes, container shape (plain dict / paginated `{count, next, previous, results}` / plain list), and the documented top-level fields and types, cross-checked against the web (`app/frontend/src/services`) and mobile (`app/mobile/src/services`) service files
- Deterministic fixtures built in `setUpTestData` (one logged-in author, a peer, an approved region/ingredient/unit, a published recipe + story, a heritage group with a journey step and a membership, a cultural fact, a cultural-content card, a cultural event, a notification); no dependency on any seed command, and shape-only assertions so the parallel seed/data PRs cannot break it
- Module docstring maps each #535 endpoint group to the test class that covers it

## Test plan
- [x] `cd app/backend && python manage.py test apps.common.tests_api_contract` (39 tests, green)
- [x] `python manage.py test` (full suite, 769 tests, green)
- [x] `python manage.py makemigrations --check --dry-run` (no changes)

## Notes
- Part A (API contract) of #535, automated. Part B (cross-platform E2E flows on the web and mobile UI) was walked through manually by the team.
- No backend contract bugs surfaced: every endpoint returns the shape the web/mobile clients expect. The only mismatches are between the *approximate* paths written in the #535 checklist and the real routes, which the suite hits by their real paths:
  - Auth lives under `/api/auth/...` (`/api/auth/register/`, `/api/auth/login/`, `/api/auth/refresh/`, `/api/auth/logout/`) and `GET /api/users/me/`, not the bare `/api/register/` etc. in the issue text.
  - Map endpoints are `/api/map/regions/`, `/api/map/regions/<id>/`, `/api/map/regions/<id>/content/`, `/api/map/discover/`, not `/api/map-discovery/regions/`.
  - There is no `/api/explore/events/` route; the "explore" surface is `GET /api/recommendations/?surface=explore` and events are `GET /api/cultural-events/` (both covered). Likewise there is no separate cultural-content "categories" endpoint; `GET /api/cultural-content/daily/`, `GET /api/cultural-events/`, `GET /api/cultural-facts/` (+ `/random/`), `GET /api/heritage-groups/`, and `GET /api/heritage-journey-steps/` are covered instead.
- Follow-up (minor, out of scope here): the recipe and story list views trigger an `UnorderedObjectListWarning` because their base querysets are unordered; adding an explicit `order_by` on those viewsets would silence it. Not a contract change.

Closes #535.

### DIFF
--- a/app/backend/apps/common/tests_api_contract.py
+++ b/app/backend/apps/common/tests_api_contract.py
@@ -1,0 +1,617 @@
+"""API contract verification suite (#535, Part A).
+
+Walks every endpoint group named in #535 and asserts the *response contract*:
+the HTTP status code, the response container shape (plain ``dict`` /
+DRF-paginated ``{count, next, previous, results}`` / plain ``list``), and the
+documented top-level fields with plausible types. It deliberately does not
+assert exact values or row counts, so seed/data PRs landing in parallel cannot
+break it. The field names asserted here are cross-checked against the web
+(``app/frontend/src/services``) and mobile (``app/mobile/src/services``) client
+service files so the suite encodes the contract those clients rely on.
+
+Coverage map (issue #535 endpoint group -> test class):
+
+  Auth ............... AuthContractTests
+  Recipes ............ RecipeContractTests
+  Stories ............ StoryContractTests
+  Search ............. SearchContractTests
+  Map ................ MapContractTests
+  Messaging .......... MessagingContractTests
+  Notifications ...... NotificationContractTests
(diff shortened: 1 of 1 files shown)


roboticrustacean (approved):
great work!

Issue #393, opened 2026-04-18
#TD-02 — Session / timeout / state management refactor


laternis:
## Description
Unify auth/session state across web and mobile: single source of truth, consistent timeout handling, clean sign-out propagation.

## Motivation / Context
Current implementation has divergent state between clients, causing subtle bugs.

## Deliverables
- Shared session state contract
- Web + mobile adoption
- Documented timeout / sign-out flow

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- MVP review notes

PR #751, opened 2026-05-12
fix(backend): login flow reliability, regression tests and root-cause note (#392)

### LINKED ISSUE #392: #TD-01 — Login flow reliability
## Description
Reproduce the login failures observed in the demo, fix the root cause, and add regression tests covering expired tokens, empty responses, and network retries.

## Motivation / Context
Multiple demo attendees hit login errors; blocks onboarding and M4 dependencies.

## Deliverables
- Root-cause write-up
- Fix + regression tests

## Acceptance Criteria

## References
- MVP demo feedback

laternis:
## Summary
- Add apps/users/tests_login_reliability.py: login happy path, bad/missing/empty/garbage/non-object request bodies all return a clean 4xx (no 500, no stack trace), wrong-method 405, token/refresh with valid/expired/malformed/blacklisted refresh, protected endpoint returns 401 token_not_valid for expired and garbage access tokens (not 403, not 500). A couple of access-token cases overlap tests_session.py (#393) on purpose so the login-reliability contract reads on its own.
- Root-cause note (docs/td-01-login-reliability.md): backend login path end to end, which conditions yield which status codes, backend vs client responsibility, conclusion that no backend bug was found and the remaining demo-login flakiness is the mobile client not handling 401 token_not_valid.
- No backend fix was needed; the login/refresh/logout/protected-endpoint paths already return correct, well-shaped responses for every failure mode the demo exposed. The value here is the regression suite plus the write-up.

## Test plan
- [x] cd app/backend && python manage.py test apps.users (87 + 16 new, green)
- [x] python manage.py test (full suite, 839 tests, green)
- [x] python manage.py makemigrations --check --dry-run (no changes detected)

## Notes
- Backend slice of #TD-01. The remaining demo-login flakiness is the mobile client not handling 401 token_not_valid, tracked in #405; the web reliability items belong to the frontend issues. Login is currently unthrottled; rate limiting is noted as a possible follow-up but is out of scope here.

Closes #392.

### DIFF
--- a/app/backend/apps/users/tests_login_reliability.py
+++ b/app/backend/apps/users/tests_login_reliability.py
@@ -1,0 +1,186 @@
+"""Login flow reliability regression tests (#392, #TD-01).
+
+Multiple demo attendees hit login errors during the MVP demo. This suite pins
+down the backend half of the login contract that the web and mobile clients
+retry against:
+
+* the login (token-obtain) endpoint returns a clean, well-shaped 4xx for bad
+  credentials, missing fields, an empty body, and a garbage body, and never a
+  500 with a leaked stack trace;
+* ``auth/token/refresh/`` behaves correctly for a valid, an expired, a
+  malformed, and a blacklisted refresh token;
+* a protected endpoint answers ``401 {"code": "token_not_valid"}`` (not 403,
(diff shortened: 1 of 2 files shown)


Emirhan777 (approved):
This fixes look good. Thank you.

PR #754, opened 2026-05-12
feat(mobile): story pins on region map detail (#731)

CaganIslam:
## Summary
RegionMapDetailScreen now plots stories alongside recipes, using `/api/stories/?region={id}` (exposed by backend #730). Recipe pins keep the green default; story pins use a mustard circular badge with a 📖 glyph and a `STORY · {title}` callout, so the two layers stay visually distinct. `fitToCoordinates` now spans both pin sets, and the header card shows `📖 N stories · 📍 M recipes`. New service helper `fetchRegionStories(regionId)` walks DRF pagination and coerces lat/lng defensively (DecimalField → string).

If a region has no story rows with lat/lng set yet, the screen silently falls back to the prior recipe-only view — no broken empty state. Couldn't verify real seeded data: backend wasn't running on localhost:8000 from this worktree, so the story layer's presence depends on whether #732-adjacent seeding has landed.

Closes #731

## Test plan
- [ ] Open MapDiscovery → tap a region pin → RegionMapDetailScreen
- [ ] Recipe pins still render in green and open RecipeDetail on callout tap
- [ ] If the region has stories with coords, mustard 📖 markers appear; callout shows `STORY · {title}` and tap opens StoryDetail
- [ ] Map fits both recipe and story pins on initial load
- [ ] Header card shows the `📖 N stories · 📍 M recipes` line
- [ ] Region with zero located stories: screen behaves exactly as before (no error/empty state)

### DIFF
--- a/app/mobile/src/screens/RegionMapDetailScreen.tsx
+++ b/app/mobile/src/screens/RegionMapDetailScreen.tsx
@@ -1,14 +1,16 @@
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { FlatList, Image, Pressable, StyleSheet, Text, View } from 'react-native';
-import MapView, { Marker } from 'react-native-maps';
+import MapView, { Callout, Marker } from 'react-native-maps';
 import { SafeAreaView } from 'react-native-safe-area-context';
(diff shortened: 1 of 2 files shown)


mustafaocakxyz (approved):
LGTM, nice backend integration

PR #761, opened 2026-05-12
feat(mobile): region-centric heritage map (#667)

CaganIslam:
## Summary
- Cluster heritage members by region and render one numbered pin per region (size scales 32/44/56 by count); top region (most members, alpha tiebreak) uses accentGreen, others accentMustard.
- Map fits all region pins then animate-centers on the top region; polylines and centroid pin removed.
- Bottom sheet (custom inline panel pinned above safe-area inset) lists each region with member rows (title + RECIPE/STORY chip); top region expanded by default, others collapsible. Tapping a marker expands+scrolls to that section.
- Members without region or coords surface as an "Unmapped" section at the end of the sheet.
- Defensive `coordsForRegion` fallback preserved; empty state copy unchanged when no member has coords.

## Sanity check
Backend `GET /api/heritage-groups/1/` ("Sarma / Dolma"): 4 members across 3 regions (Black Sea ×2, Southeastern Anatolia ×1, Aegean ×1). Top region = Black Sea.

## Implementation notes
- Did not reuse `RegionDetailSheet` — that component is coupled to `fetchRegionContent(regionId)` and tabbed recipes/stories, which doesn't match the heritage-member shape. Built a lightweight in-screen sheet to keep the data path local.
- Size scaling: discrete 3-tier (1 / 2-4 / 5+). Collapse behavior: top expanded by default, the rest collapsed with tap-to-expand headers.
- Used `useSafeAreaInsets()` for the bottom inset (this screen is in the root stack, not under the tab bar).

## Test plan
- [ ] Open a heritage group from Heritage screen, tap "View map".
- [ ] Verify one pin per region; pin sizes vary with member count; top region pin is green.
- [ ] Map fits all pins on load and centers on the top region.
- [ ] Tap a non-top pin → bottom sheet scrolls and expands that region.
- [ ] Tap a member row → opens RecipeDetail or StoryDetail accordingly.
- [ ] Group with only one mappable region → still renders one pin and the sheet.
- [ ] Group with zero mappable members → existing empty-state copy.
- [ ] "Unmapped" section appears when at least one member has no region/coords.

Closes #667

### DIFF
--- a/app/mobile/src/screens/HeritageMapScreen.tsx
+++ b/app/mobile/src/screens/HeritageMapScreen.tsx
@@ -50,6 +50,74 @@ function markerSizeForCount(count: number): number {
   return 32;
 }
 
+/** Just the colored dot — hardware-textured + offscreen-composited so Android
+ * properly rounds the corners when it snapshots the view into the marker
+ * bitmap (without this, borderRadius is silently dropped and the marker
+ * renders as a coloured square). */
+function HeritageRegionDot({ size, isTop }: { size: number; isTop: boolean }) {
+  return (
+    <View
+      collapsable={false}
+      renderToHardwareTextureAndroid
+      needsOffscreenAlphaCompositing
+      style={{
+        width: size + 8,
+        height: size + 8,
+        alignItems: 'center',
+        justifyContent: 'center',
+        backgroundColor: 'transparent',
+      }}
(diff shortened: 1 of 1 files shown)


ErenCanOzkaya (approved):
LGTM

PR #717, opened 2026-05-11
feat(frontend): milestone 8 web — heritage, journey, did-you-know, calendar

ErenCanOzkaya:
## Summary

  Closes Milestone 8 web frontend work for the four ready issues. Backend was already merged on `main` for these (#499, #511, #522, #528 closed); this PR is the matching web layer.

  - **#500 Heritage Badge, Page & Map**
    - `HeritageBadge` component appears on `RecipeDetail` and `StoryDetail` when the API returns a `heritage_group`, links to `/heritage/<id>`
    - `HeritagePage` at `/heritage/:id`: header, "Show Heritage Map" CTA, members grid (recipes + stories mixed), graceful empty/error states
    - `HeritageMapPage` at `/heritage/:id/map`: Leaflet view with the central 🏛 marker at the geographic midpoint of member coordinates, dashed polylines from midpoint to each located
   member, color-coded `CircleMarker` per recipe/story, click-through to detail, "← Back to Heritage Page" link
    - Members without `latitude/longitude` are filtered out of the map; empty state renders when none have coords
    - `computeHeritageMidpoint` / `locatableMembers` extracted as a pure util so the math is unit-tested without Leaflet

  - **#515 Heritage Journey**
    - `HeritageJourneySection` timeline component rendered inside `HeritagePage`
    - Steps sorted by `order` ascending; era badge only when present; prop is never mutated

  - **#517 Cultural Context Cards — "Did You Know?"**
    - `CulturalFactCard` reusable card with `target="_blank"` + `rel="noopener noreferrer"` on source links
    - Facts grid on `HeritagePage` (filtered by heritage group)
    - Facts list on `RecipeDetailPage` when the recipe has a `heritage_group` (no API call for unaffiliated recipes)
    - `RandomCulturalFact` widget on `HomePage` — silently renders nothing on null / 404 / error

  - **#527 Seasonal & Ritual Calendar**
    - `CalendarPage` at `/calendar`: 12-month grid + dedicated "Lunar-Anchored Events" panel for `lunar:*` rules
    - Month + region filters drive a fresh `/api/cultural-events/?month=MM&region=<id>` query
    - Click any event card → side detail panel with description, region, and linked recipes (`Link` to `/recipes/<id>`)
    - New "Calendar" entry in the public navbar between Explore and the user menu

  ### Service / util layer

  - `heritageService.js` — `fetchHeritageGroups`, `fetchHeritageGroup` (DRF pagination unwrap)
[...]

PR #762, opened 2026-05-12
fix(frontend): post-M8 review feedback — null members guard and HomePage test stability

ErenCanOzkaya:
## Summary

Follow-up to #717 (Milestone 8 web). Addresses two concrete review comments:

- **HeritagePage** — `group.members.length` now tolerates an API response where `members` is `null`/missing (guards with `!group.members || group.members.length === 0`). Prevents a runtime crash on the empty path.
- **HomePage test "populates region chips from API"** — swap `waitFor` for `findByRole` so the assertion races with a single retried query instead of polling alongside the parallel `RandomCulturalFact` and `DailyCulturalSection` effects. Removes the `act()` warnings and makes the test resilient to CI scheduling jitter.

## Test plan

- [ ] `npm test` in `app/frontend/` — full suite green (504/504 locally)
- [ ] HeritagePage still renders the empty state when a group has no members
- [ ] HomePage tests no longer surface `act()` warnings tied to RandomCulturalFact

### DIFF
--- a/app/frontend/src/__tests__/HomePage.test.jsx
+++ b/app/frontend/src/__tests__/HomePage.test.jsx
@@ -53,10 +53,11 @@ describe('HomePage', () => {
 
   it('populates region chips from API', async () => {
     renderPage();
-    await waitFor(() => {
-      expect(screen.getByRole('button', { name: 'Aegean' })).toBeInTheDocument();
-      expect(screen.getByRole('button', { name: 'Mediterranean' })).toBeInTheDocument();
-    });
+    // Use findByRole so the assertion races with a single retried query
+    // instead of waitFor polling — the latter interleaves with async state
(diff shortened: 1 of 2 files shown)


Daglar1500 (approved):
Frontend CI geçiyor. HeritagePage null guard minimal ve doğru —  olmadığında crash'i önlüyor. HomePage testi  ile flaky davranış giderilmiş, yaklaşım standart. Backend CI yine timeout (runner sorunu, bu PR'a özgü değil). Merge edilebilir.

Daglar1500 (approved):
Frontend CI geçiyor. HeritagePage null guard minimal ve doğru. HomePage testi findByRole ile flaky davranış giderilmis. Backend CI yine timeout (runner sorunu, bu PR'a ozgu degil). Merge edilebilir.

PR #763, opened 2026-05-12
fix(frontend): add .env.test — fixes 79 failing service tests

Daglar1500:
## Summary

**Root cause:** `.env.test` was missing. Jest also loads `.env` (CRA default behaviour), which has `REACT_APP_USE_MOCK=true`. This caused every service function to take the mock branch in the test environment — `apiClient` was never called. Service tests that assert `apiClient.get/.post/.patch` saw 0 calls.

**Affected:** 12 service test suites, 79 tests (authService, recipeService, storyService, commentService, etc.)

**Fix:** One line — added `app/frontend/.env.test`:
```
REACT_APP_USE_MOCK=false
```

The project convention requires this file to exist so that service tests exercise the real `apiClient` code path in CI. The file was never created.

**Result:** 66 suites, 513 tests passing (previously 12 suites / 79 tests failing).

## Test plan
- [ ] `CI=true npm test` → 513 passed, 0 failed

### DIFF
--- a/app/backend/apps/common/tests_e2e_scenarios.py
+++ b/app/backend/apps/common/tests_e2e_scenarios.py
@@ -1,0 +1,494 @@
+"""End-to-end flow scenarios (#397, #QA-01 backend slice).
+
+Multi-step, server-side flows that walk several endpoints in sequence the way
+a real user would, asserting that state carries correctly from one step to the
+next. This is the layer above the per-endpoint contract suite (#535,
+``apps/common/tests_api_contract.py``): #535 pins each endpoint's response
+shape, this module pins the cross-endpoint behaviour (a notification raised by
+one call is visible to the other party on a later call, an inbox reflects a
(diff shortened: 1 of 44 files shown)


ErenCanOzkaya (commented):
This PR is **obsolete** — `app/frontend/.env.test` was already added on `main` by #719 (commit `a0dd75a`, merged into PR #719). The main version is actually a superset of what this PR adds:

```
# main:
REACT_APP_API_URL=http://localhost:8000
REACT_APP_USE_MOCK=false

# this PR:
REACT_APP_USE_MOCK=false
```

That's why `mergeable: CONFLICTING` — applying this PR would drop the `REACT_APP_API_URL` line that main now has.

**Please close this PR.** No further action needed — service tests on main already pass against `REACT_APP_USE_MOCK=false`.

CaganIslam (approved):
Minimal fix, 1 line, all 513 tests pass. LGTM

Issue #590, opened 2026-05-10
feat(frontend): add I Tried This, Add to My Passport, and Save to Passport buttons


laternis:
## Summary
Add explicit passport action buttons to recipe and story detail pages.

## Scope
**RecipeDetailPage**
- `I Tried This` button → `POST /api/passport/recipes/:id/try/`
- `Add to My Passport` button → save without "tried" status (parameter on the same endpoint, or a separate save route)

**StoryDetailPage**
- `Save to Passport` button → `POST /api/passport/stories/:id/save/`

Buttons visible only to authenticated users. After action, the button switches to a filled/active state and is disabled for re-trigger (idempotency is enforced server-side).

## Files Affected
- `app/frontend/src/pages/RecipeDetailPage.jsx`
- `app/frontend/src/pages/RecipeDetailPage.css`
- `app/frontend/src/pages/StoryDetailPage.jsx`
- `app/frontend/src/pages/StoryDetailPage.css`
- `app/frontend/src/services/passportService.js` (`addRecipeToPassport`, `saveStoryToPassport`)

## Acceptance Criteria
- [ ] Anonymous users do not see the buttons
- [ ] First click triggers the API call; success toast or inline confirmation shown
- [ ] Active state persists across navigation (response stays accurate)
- [ ] Failure shows an error toast, button remains clickable
- [ ] Tests cover visibility, success, and failure cases

## Depends On
- Backend #584 (passport action endpoints)

## Notes
Per the spec, "Add to Passport" vs "I Tried This" are two distinct affordances. If usability tests show overlap, the team can collapse them later.

PR #764, opened 2026-05-12
feat(frontend): add I Tried This, Add to Passport, Save to Passport buttons (#590)

Daglar1500:
## Summary

passportService: `tryRecipe`, `addRecipeToPassport`, `saveStoryToPassport` fonksiyonları eklendi. Mock mode'da anında `{ status: 'ok' }` döner, gerçek modda backend endpoint'lerine POST atar.

RecipeDetailPage: "I Tried This" ve "Add to Passport" butonları eklendi. Sadece giriş yapmış kullanıcılara görünür. Başarı sonrası buton aktif state'e geçer ve disable edilir. Hata durumunda buton tekrar tıklanabilir kalır.

StoryDetailPage: "Save to Passport" butonu eklendi. Aynı davranış.

## Test plan
- [ ] `/recipes/:id` — giriş yapmadan: butonlar görünmez
- [ ] `/recipes/:id` — giriş yaparak: "I Tried This" + "Add to Passport" görünür
- [ ] "I Tried This" tıkla → "✓ I Tried This" olur, disable edilir
- [ ] "Add to Passport" tıkla → "✓ In Passport" olur, disable edilir
- [ ] `/stories/:id` — giriş yaparak: "Save to Passport" görünür, tıkla → "✓ Saved to Passport"
- [ ] API hatası → hata mesajı çıkar, buton tekrar tıklanabilir
- [ ] `CI=true npm test` → passportActions: 10/10 pass

### DIFF
--- a/app/frontend/src/__tests__/passportActions.test.jsx
+++ b/app/frontend/src/__tests__/passportActions.test.jsx
@@ -82,22 +82,19 @@ beforeEach(() => {
   culturalFactService.fetchCulturalFacts.mockResolvedValue([]);
   storyService.fetchStory.mockResolvedValue(mockStory);
   passportService.tryRecipe.mockResolvedValue({ status: 'ok' });
-  passportService.addRecipeToPassport.mockResolvedValue({ status: 'ok' });
   passportService.saveStoryToPassport.mockResolvedValue({ status: 'ok' });
 });
 
 describe('Passport action buttons — RecipeDetailPage', () => {
-  it('does not show passport buttons when logged out', async () => {
+  it('does not show I Tried This button when logged out', async () => {
     renderRecipePage(null);
     await waitFor(() => expect(screen.getByText('Baklava')).toBeInTheDocument());
     expect(screen.queryByRole('button', { name: /i tried this/i })).not.toBeInTheDocument();
-    expect(screen.queryByRole('button', { name: /add to passport/i })).not.toBeInTheDocument();
   });
 
-  it('shows both passport buttons when logged in', async () => {
+  it('shows I Tried This button when logged in', async () => {
     renderRecipePage(loggedInUser);
     await waitFor(() => expect(screen.getByRole('button', { name: /i tried this/i })).toBeInTheDocument());
-    expect(screen.getByRole('button', { name: /add to passport/i })).toBeInTheDocument();
   });
 
   it('clicking I Tried This calls tryRecipe and disables the button', async () => {
@@ -106,14 +103,6 @@ describe('Passport action buttons — RecipeDetailPage', () => {
     await userEvent.click(screen.getByRole('button', { name: /i tried this/i }));
     expect(passportService.tryRecipe).toHaveBeenCalledWith('1');
     await waitFor(() => expect(screen.getByRole('button', { name: /✓ i tried this/i })).toBeDisabled());
-  });
-
-  it('clicking Add to Passport calls addRecipeToPassport and disables the button', async () => {
-    renderRecipePage(loggedInUser);
-    await waitFor(() => expect(screen.getByRole('button', { name: /add to passport/i })).toBeInTheDocument());
-    await userEvent.click(screen.getByRole('button', { name: /add to passport/i }));
-    expect(passportService.addRecipeToPassport).toHaveBeenCalledWith('1');
-    await waitFor(() => expect(screen.getByRole('button', { name: /✓ in passport/i })).toBeDisabled());
   });
 
   it('tryRecipe failure shows error, button stays clickable', async () => {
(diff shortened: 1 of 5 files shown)


ErenCanOzkaya (commented):
The button UX, optimistic active state, and the `passportActions.test.jsx` coverage are all solid. Two blockers before this can land:

## Blocking #1 — main rebase deletes ChatTray (#719)

This branch was forked off `feat/frontend/chat-tray-716` **before** #719 merged. The diff against current main now shows the entire #716 chat tray feature being deleted:

```
- app/frontend/.env.test                          (–2)
- app/frontend/Dockerfile                         (–2 ChatTray env args)
- app/frontend/src/App.js                         (–2 ChatTray render)
- app/frontend/src/__tests__/ChatContext.test.jsx (–89)
- app/frontend/src/__tests__/ChatTray.test.jsx    (–180)
- app/frontend/src/components/ChatTray.css        (–380)
- app/frontend/src/components/ChatTray.jsx        (–233)
- app/frontend/src/context/ChatContext.jsx        (–55)
- app/frontend/src/index.js                       (–4 ChatProvider wrap)
- docker-compose.yml                              (–1 USE_MOCK arg)
- app/frontend/src/components/Navbar.jsx          (+7 — Inbox link re-added)
```

That's a full regression of #716. Please rebase against `main` and accept main's side on every conflict in that list (basically keep #719 untouched). Only the new files for #590 — `passportService.js` additions, the two detail-page button blocks, and `passportActions.test.jsx` — should remain in the diff after rebase.

## Blocking #2 — `addRecipeToPassport` calls a non-existent endpoint

`passportService.js` posts to `/api/passport/recipes/<id>/save/`, but that route does not exist in the backend. The only recipe action route registered in `apps/passport/urls.py` is:

```python
path('passport/recipes/<str:recipe_id>/try/', TryRecipeView.as_view(), …)
path('passport/stories/<str:story_id>/save/', SaveStoryView.as_view(), …)
```

Verified against the running stack: `POST /api/passport/recipes/1/save/` returns no route match. The backend deliberately collapses "tried" and "saved" into the single `/try/` endpoint — the action body of `TryRecipeView` already updates the passport stamp + advances quests + emits a timeline event in one shot, so there isn't a separate "save without try" action.

Two ways forward, your call:
- **(a)** Drop the `Add to Passport` button and ship only `I Tried This`. The backend doesn't model save-without-try as a separate state today, and #590's AC says "If usability tests show overlap, the team can collapse them later" — this is that moment.
- **(b)** Keep both buttons but have them call the same `/try/` endpoint, treat them as visually-distinct shortcuts to the same idempotent action. Less honest UX-wise.

I'd recommend (a). The story-side `saveStoryToPassport` is fine — that endpoint exists.

## Minor

- Optimistic active state stays after a successful POST, but if the user navigates away and back the page reloads `recipe.heritage_group` etc. without knowing the recipe is now tried. The `TryRecipeView` response payload already includes the updated `culture_summaries` and `stamps` — would be nice to hydrate that into a `tried_recipes` derived flag on the recipe-detail page later (out of scope for this PR, just flagging).
- `passportActions.test.jsx` tests cover the happy path and the unauthenticated case well. Could you add one for the error branch (POST rejects) so we lock in "button stays enabled on failure"? Looks like the implementation handles it, but the test would prove it.

Once rebased + endpoint resolved I'll re-review.

CaganIslam (approved):
Clean implementation with solid test coverage, all 523 tests pass. LGTM

ErenCanOzkaya (approved):
You were right and my first review was wrong on Blocker 1 — apologies for the noise.

I checked with `git diff origin/main..feat/frontend/passport-actions-590` (two-dot diff), which shows the difference between branch tips and made the ChatTray files look like deletions. The real merge uses a **three-way merge** against the merge-base (`11dd10e`, PR #673 merge commit). Since this branch never edited or deleted any ChatTray file, those additions on `main` (introduced after the branch was cut) are preserved as-is in the merge result.

Just simulated it locally: `git merge origin/main` into a copy of this branch leaves `ChatTray.jsx`, `ChatTray.css`, `ChatContext.jsx`, `.env.test`, the two ChatTray test files, and the Dockerfile/docker-compose ENV bits all intact. Full test suite reports **542 passed, 542 total** on the post-merge tree, including the eight new `passportActions` tests.

**Blocker 2** is properly resolved — `addRecipeToPassport` and the `/recipes/<id>/save/` POST are gone from `passportService.js`, `RecipeDetailPage.jsx`, and the test file. The only recipe action is `tryRecipe` → `/api/passport/recipes/<id>/try/`, which matches the backend route in `apps/passport/urls.py`.

Looks good — ready to merge.

ErenCanOzkaya (approved):
Re-approving after the rebase — clean diff against current main, full suite green locally.

PR #765, opened 2026-05-12
feat(frontend): Cultural Passport tabs — PassportCover, StatsBar, Stamps, Cultures, Map, Timeline, Quests (#591-#597)

Daglar1500:
## Summary

UserProfilePage'deki passport placeholder'ı, 5 sekmelik tam Cultural Passport bölümüyle değiştirdi.

passportService: 4 stub endpoint → gerçek API çağrıları + `REACT_APP_USE_MOCK=true` için zengin mock data.

**Yeni bileşenler** (`src/components/passport/`):
- `PassportCover` — `culturalCalendar.resolveTheme()` ile takvim/seviye bazlı tema seçimi
- `PassportStatsBar` — 5 metrik + seviye rozeti (Bronze → Master)
- `StampGrid` + `StampCard` — kategori gruplandırması, nadir renk sınırları, progress bar
- `CultureGrid` + `CultureDetailPanel` — kültür kartları, tıkla → inline detay paneli
- `PassportMap` — SVG dünya haritası, bölge renkleri `passportMapColors.getRegionFill()` ile
- `PassportTimeline` — kronolojik olay akışı, recipe/story linkleri
- `QuestList` — aktif/tamamlanmış görevler, progress bar, deadline

**Yeni utils** (`src/utils/`):
- `culturalCalendar.js` — `resolveTheme(profile, level, date)`
- `passportMapColors.js` — `getRegionFill(culture)`

UserProfilePage: 4 passport endpoint'i `Promise.allSettled` ile paralel fetch eder; biri başarısız olursa diğerleri etkilenmez.

## Test plan
- [ ] `CI=true npm test` → 588 passed, 0 failed
- [ ] `REACT_APP_USE_MOCK=true npm start` → `/users/demo_chef` → PassportCover görünür, 5 sekme çalışır
- [ ] Stamps sekmesi → kategori başlıkları, renkli kenarlıklar, kilitli stamp
- [ ] Cultures sekmesi → kart tıkla → detay paneli açılır, kapat → kapanır
- [ ] Timeline sekmesi → recipe/story linkleri doğru hedefleri gösteriyor
- [ ] Quests sekmesi → aktif/tamamlanmış bölümleri, deadline görünür
- [ ] Giriş yapılmamış ziyaretçi → sekmeler görünür ama edit butonu yok

PR #769, opened 2026-05-12
fix(mobile): heritage map default pins + scrollable sheet

CaganIslam:
## Summary
- Replaces the custom round-view region markers with the native platform pin (`Marker pinColor` prop). The previous custom view rendered as a square on this Android setup despite every combination of `renderToHardwareTextureAndroid` / `needsOffscreenAlphaCompositing` / `overflow:hidden` / dual `tracksViewChanges` / unicode glyph. Native pins are bitmap-safe and instantly readable — top region in `accentGreen`, others in `accentMustard`. Size tiers dropped.
- Raises sheet `maxHeight` from 55% → 75% and adds `nestedScrollEnabled` to the inner ScrollView so the region list actually scrolls when content overflows on Android.

## Test plan
- [ ] Open a heritage group with multiple regions (e.g. `Sarma / Dolma` has 6) — pins are platform-native teardrops in green (top region) and mustard
- [ ] Sheet content beyond ~3 expanded regions — drag to scroll, list moves
- [ ] Tap a marker — that region expands in the sheet and scrolls into view

### DIFF
--- a/app/mobile/src/screens/HeritageMapScreen.tsx
+++ b/app/mobile/src/screens/HeritageMapScreen.tsx
@@ -257,6 +257,7 @@ export default function HeritageMapScreen({ route, navigation }: Props) {
   return (
     <SafeAreaView style={styles.safe} edges={['top', 'left', 'right']}>
       <View style={styles.fill}>
+        <View style={styles.mapContainer}>
         <MapView
           ref={mapRef}
           style={styles.map}
@@ -296,6 +297,7 @@ export default function HeritageMapScreen({ route, navigation }: Props) {
             {clusters.length} region{clusters.length === 1 ? '' : 's'} · {totalMembers} member
             {totalMembers === 1 ? '' : 's'}
           </Text>
+        </View>
         </View>
 
         <View style={[styles.sheet, { paddingBottom: sheetBottomInset + 12 }]}>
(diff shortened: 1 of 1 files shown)


ErenCanOzkaya (approved):
Mobile change reads cleanly. Dropping the custom round-view marker in favour of the platform `pinColor` API is the right call given the Android `borderRadius` / Marker-bitmap interaction — the `tracksViewChanges` + hardware-texture + offscreen-composition trifecta was a workaround for a snapshot quirk that `pinColor` sidesteps entirely. Cluster size info is still surfaced through the `description` ("X members") and through the legend, so removing the three-tier size scale is acceptable; if size-as-emphasis is ever wanted back, we can layer it via a label or a tiny custom callout without re-introducing the round-view path.

The sheet refactor (`flex: 0` + `flexBasis: '50%'` + `maxHeight: 60%` inside the new `mapContainer` flex parent) is the right shape for letting `ScrollView` claim its own gestures on Android — `position: absolute` over `MapView` swallowed touches before. `nestedScrollEnabled` + `keyboardShouldPersistTaps="handled"` are the right knobs.

## CI

I see the red checks; concur with your heads-up note — the `UserProfilePage.test.jsx` failure is a `getPassport` mock gap that arrived on `main` with #765 and is unrelated to this PR's mobile-only scope. I also locally verified the diff: the 28 frontend deletions visible in the two-dot diff (`git diff origin/main..pr-branch`) are a merge-base artifact, not actual deletions in this branch. Simulated `git merge origin/main` into this PR: all Passport components, utils, and tests on `main` are preserved post-merge, and your `pinColor` change applies cleanly. Safe to merge.

Approving — the UserProfilePage test fix is in flight separately and will turn CI green once it lands.

PR #773, opened 2026-05-12
fix(mobile): dedupe paginated story/recipe lists by id

CaganIslam:
## Summary
Defensive Map-by-id dedupe in `fetchStoriesList` and `fetchRecipesList` after the pagination walk concatenates pages. Backend occasionally returns the same row on consecutive pages when the queryset ordering has no PK tiebreaker — surfacing as the 'Encountered two children with the same key, \`31\`' console warning on the HomeScreen stories rail.

Workaround for backend #770 (Story/Recipe pagination ordering instability).

## Test plan
- [ ] Open Home — no more duplicate-key warning toast at the bottom
- [ ] Stories rail and 'More recipes' rail still render the expected items
- [ ] No regression in author-filtered profile views (own profile Recipes/Stories tabs)

### DIFF
--- a/app/mobile/src/services/recipeService.ts
+++ b/app/mobile/src/services/recipeService.ts
@@ -155,7 +155,15 @@ export async function fetchRecipesList(filter?: { author?: number | string }): P
       break;
     }
   }
-  return collected.map((r) => {
+  // Defensive dedupe by id — see backend #770 (pagination ordering instability).
+  const byId = new Map<string | number, any>();
+  const idLess: any[] = [];
+  for (const r of collected) {
+    if (r && r.id != null) byId.set(r.id, r);
+    else idLess.push(r);
+  }
+  const deduped = [...byId.values(), ...idLess];
+  return deduped.map((r) => {
     const reg = r.region;
     const regionLabel =
       reg == null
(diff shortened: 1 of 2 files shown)


ErenCanOzkaya (approved):
Defensive dedupe is the right move while backend #770 (pagination ordering instability) is open. Map-based id grouping with last-write-wins is the safest shape — callers downstream that already key on `id` will get the same object identity per id within a single fetch.

A couple of small observations:

- `id == null` items end up in the `idLess` array and are appended at the end. That preserves any rows the API sends without a primary key (shouldn't happen, but the guard is cheap and the alternative — dropping them — would lose data). Good defensive call.
- Map ordering is insertion-ordered in modern JS engines, so the merged list preserves the API's intended sequence for the unique ids encountered, with later duplicates overwriting earlier ones. If the backend ever uses page ordering as a signal (e.g., "newest first"), the dedupe keeps the latest seen version, which matches what consumers expect.

Both `recipeService` and `storyService` end up with the same shape; could be DRY'd into a tiny helper later if a third paginated service shows up.

## CI

Concur with your CI heads-up — the `UserProfilePage.test.jsx` failures are a known `getPassport` mock gap that landed with #765 on `main`, completely outside this PR's mobile scope. This PR only modifies `app/mobile/src/services/*.ts`. Once the UserProfilePage fix lands, a rerun here will go green.

Approving — safe to merge as-is, fully unblocked by the in-flight UserProfilePage test fix.

PR #774, opened 2026-05-12
feat(frontend): apply animated food grid background to every page (#772)

ErenCanOzkaya:
Closes #772.

## Summary

Promote the animated food-image grid (`GridMotion`) from auth-only to a site-wide background. Page content stays on top of a darker overlay so dense screens (calendar grid, recipe lists, heritage map) remain readable.

- New `AppBackground` component owns the food image list and renders the two background layers (`.app-bg` grid + `.app-bg-filter` overlay). Mounted unconditionally from `App.js` — the `isAuthPage` guard is gone.
- Overlay darkened from `rgba(0,0,0,0.4)` → `rgba(0,0,0,0.7)` to keep dense content legible.
- `.app-bg` / `.app-bg-filter` styles moved into `components/AppBackground.css` (out of `LoginPage.css`, where they no longer belong).
- `.page-wrapper` gets `position: relative; z-index: 2;` so every route's content sits above the overlay (the existing auth-page rule already had its own z-index).
- `GridMotion` respects `prefers-reduced-motion: reduce` — users who opt out get a static grid (no GSAP timeline, no scrolling).
- Both background layers are marked `aria-hidden="true"` so the grid does not pollute the accessibility tree.
- Two Jest tests cover the mount and a11y attributes for `AppBackground`.

## Test plan

- [ ] `CI=true npm test` — 582 tests pass (8 pre-existing failures in `UserProfilePage.test.jsx` are unrelated — `getPassport` is not mocked there since PR #765; tracked separately)
- [ ] `npm run build` — production build succeeds
- [ ] `http://localhost/` — grid + dark overlay visible behind home page, search/filter chips remain readable
- [ ] `/recipes`, `/stories`, `/calendar`, `/heritage/<id>`, `/users/<username>` — same check, no z-index regressions
- [ ] `/heritage/<id>/map` — Leaflet sits above the overlay, pins still clickable
- [ ] `/login` and `/register` — auth pages still look correct, no double-overlay
- [ ] Chat tray (bottom-left, z-index 900) and navbar (z-index 50) sit above the overlay
- [ ] OS-level "Reduce motion" setting on → grid renders as a static image strip, no animation

### DIFF
--- a/app/frontend/src/App.js
+++ b/app/frontend/src/App.js
@@ -1,9 +1,9 @@
-import { Routes, Route, useLocation } from 'react-router-dom';
+import { Routes, Route } from 'react-router-dom';
 import Navbar from './components/Navbar';
 import SkipLink from './components/SkipLink';
 import ChatTray from './components/ChatTray';
 import ProtectedRoute from './components/ProtectedRoute';
-import GridMotion from './components/GridMotion';
+import AppBackground from './components/AppBackground';
 import HomePage from './pages/HomePage';
 import LoginPage from './pages/LoginPage';
 import RegisterPage from './pages/RegisterPage';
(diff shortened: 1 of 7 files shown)


CaganIslam (approved):
All tests pass, clean extraction. LGTM

PR #776, opened 2026-05-12
feat(mobile): notifications scaffolding — screen, service, bell icon entry (closes #760)

CaganIslam:
## Summary
- New `notificationService.ts` wraps the backend list, mark-read, and mark-all-read endpoints, walking DRF pagination via `nextPagePath`.
- New `NotificationsScreen.tsx` renders a FlatList with per-type icons (💬 question, ↪ reply, ⭐ rating, ❔ fallback), inline time-ago helper, unread mustard dot, pull-to-refresh, empty/error states, and a top-bar Mark-all-read pill.
- HomeScreen gains a 🔔 bell button (top-right of the Search header) with a mustard unread-count badge; tap routes to `Notifications` when signed in or `Login` otherwise.
- Navigation wired: `Notifications: undefined` added to `RootStackParamList`, screen registered in `PublicStackNavigator`.
- Backend endpoints used: `GET /api/notifications/`, `POST /api/notifications/<id>/read/`, `POST /api/notifications/read-all/`. No `unread_count/` endpoint exists yet, so we compute it client-side from the paginated list.

## Test plan
- [ ] Open Home as a signed-out user — bell renders without a badge; tap routes to Login.
- [ ] Sign in, open Home — bell shows current unread count, refreshes on focus.
- [ ] Tap the bell — Notifications screen lists items newest-first with correct icons and relative time strings.
- [ ] Pull to refresh — list reloads; unread dot appears only on `is_read: false` rows.
- [ ] Tap an unread item — row marks read optimistically and navigates to `RecipeDetail` when `recipe` is set.
- [ ] Tap "Mark all read" pill — every row flips to read; pill hides when no unread remain.
- [ ] Empty inbox shows "No notifications yet." copy.

Closes #760

### DIFF
--- a/app/mobile/src/navigation/PublicStackNavigator.tsx
+++ b/app/mobile/src/navigation/PublicStackNavigator.tsx
@@ -10,6 +10,7 @@ import InboxScreen from '../screens/InboxScreen';
 import LoginScreen from '../screens/LoginScreen';
 import MapDiscoveryScreen from '../screens/MapDiscoveryScreen';
 import MessageThreadScreen from '../screens/MessageThreadScreen';
+import NotificationsScreen from '../screens/NotificationsScreen';
 import OnboardingScreen from '../screens/OnboardingScreen';
 import RecipeCreateScreen from '../screens/RecipeCreateScreen';
 import RecipeDetailScreen from '../screens/RecipeDetailScreen';
@@ -140,6 +141,11 @@ export function PublicStackNavigator() {
         name="IngredientMigrationMap"
         component={IngredientMigrationMapScreen}
         options={{ title: 'Migration routes' }}
+      />
+      <Stack.Screen
+        name="Notifications"
+        component={NotificationsScreen}
+        options={{ title: 'Notifications' }}
       />
     </Stack.Navigator>
   );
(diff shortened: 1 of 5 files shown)


ErenCanOzkaya (approved):
Solid scaffolding work — service, screen, and bell-icon entry all hang together cleanly.

## Highlights

- `notificationService.ts` has the right defensive normalisation: `toNum`, `normalizeActor`, and `is_read === true` (so missing/null fields don't accidentally render as truthy). Pagination walked via the existing `nextPagePath` helper — consistent with `recipeService` / `storyService`.
- Inline `timeAgo` helper avoids a date-fns / dayjs dep; the bundle-size note in the comment is exactly the right justification.
- Bell badge is correctly treated as decorative — `if (!isAuthenticated) return 0` short-circuits the network call (no 401 noise on the home screen for logged-out users), and a fetch error silently degrades. Auth-aware refresh on focus is the right shape.
- `NotificationsScreen` covers the full state matrix (loading / empty / error / pull-to-refresh / mark-one-read / mark-all-read). `accessibilityRole="header"` on the title and per-row accessibility labels are nice a11y touches.
- `Notifications` registered in `RootStackParamList` and `PublicStackNavigator` — no orphan navigate.

## CI

Red `Frontend tests` check is the known `UserProfilePage.test.jsx` regression on `main` (missing `getPassport` mock — fix in flight as #780). Three-way merge simulation against current `main` is clean: zero conflicts, only the 5 mobile files in your description actually change. Approving — rerun the workflow after #780 lands and it'll go green.

PR #777, opened 2026-05-12
feat(mobile): redesign Profile tab — identity header + action list (closes #447)

CaganIslam:
## Summary
- Replaces the ProfileTab placeholder with a proper authenticated layout: tap-to-open identity header (avatar circle + @username + email) and a vertical action list.
- Action rows: My recipes, My stories, Saved recipes, Messages, Cultural preferences, Edit profile, Log out (danger pink).
- My recipes / My stories / Saved recipes all navigate to the user's UserProfile screen (deep-link to specific tabs skipped — UserProfile does not accept an `initialTab` param yet, so users tap the in-screen chip).
- Edit profile shows a "Coming soon" toast (Android `ToastAndroid`, iOS `Alert`) because no `EditProfile` route exists in `RootStackParamList`. Wiring can land once that screen exists.
- Unauthenticated state keeps Log In / Register, with a short cultural greeting above as the issue requested.

## Test plan
- [ ] Open Profile tab while signed in — header shows initial avatar, @username, email.
- [ ] Tap header — UserProfile opens for the current user.
- [ ] Tap My recipes / My stories / Saved recipes — UserProfile opens (chip selection still works inside).
- [ ] Tap Messages — Inbox opens.
- [ ] Tap Cultural preferences — Onboarding opens.
- [ ] Tap Edit profile — toast/alert "Coming soon" appears.
- [ ] Tap Log out — session ends, unauth state appears.
- [ ] Sign out and reopen Profile — greeting + Log In / Register pair render.

Closes #447

### DIFF
--- a/app/mobile/src/screens/tabs/ProfileTabScreen.tsx
+++ b/app/mobile/src/screens/tabs/ProfileTabScreen.tsx
@@ -1,78 +1,157 @@
 import React from 'react';
 import { useNavigation } from '@react-navigation/native';
-import { Pressable, StyleSheet, Text, View } from 'react-native';
+import {
+  Alert,
+  Platform,
+  Pressable,
+  ScrollView,
+  StyleSheet,
+  Text,
+  ToastAndroid,
+  View,
+} from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 import { useAuth } from '../../context/AuthContext';
 import { shadows, tokens } from '../../theme';
+
+function showComingSoon() {
+  if (Platform.OS === 'android') {
+    ToastAndroid.show('Coming soon', ToastAndroid.SHORT);
+  } else {
+    Alert.alert('Coming soon', 'Edit profile is not available yet.');
+  }
+}
+
+type ActionRowProps = {
+  icon: string;
+  label: string;
+  onPress: () => void;
+  accessibilityLabel: string;
+  danger?: boolean;
+};
+
+function ActionRow({ icon, label, onPress, accessibilityLabel, danger }: ActionRowProps) {
+  return (
+    <Pressable
+      onPress={onPress}
+      style={({ pressed }) => [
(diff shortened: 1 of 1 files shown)


ErenCanOzkaya (approved):
Clean redesign — the ActionRow pattern is the right abstraction (icon + label + chevron, with a `danger` flag for the logout row) and keeps the file easy to scan.

## Highlights

- Tap-to-open identity header navigates into `UserProfile`, mirroring the web behaviour where the navbar `@username` chip jumps to the profile page. Good consistency across platforms.
- `showComingSoon` correctly uses `ToastAndroid` on Android and `Alert` on iOS — that's the right native idiom for each platform.
- Logged-out state has both Log In and Register CTAs with proper a11y labels.
- Danger styling on the logout row (the pink) is a nice destructive-action signal without needing a confirm dialog (which `logout` itself doesn't currently prompt for — fine, just noting; we can add one later if accidental taps become a thing).
- Three "My X" rows all hit the same `openUserProfile` — the user's own profile page is where those lists are surfaced. Worth a follow-up if/when "Saved recipes" gets its own tab/view, but for now consistent with the back-end shape.

## Minor

- The deep-link `navigation.navigate('Feed', { screen: 'UserProfile', params: { userId, username } })` assumes the Feed stack is mounted. If a future refactor moves UserProfile out of Feed, every `openUserProfile` callsite breaks at runtime. Could be hoisted into a single `goToProfile(user)` helper in `useAuth` (or a `navigation/links.ts`) for the next sweep — not blocking.
- `Edit profile` showing "Coming soon" is fine for #447 scope, just confirming this isn't accidentally wired to a screen that already exists.

## CI

Red `Frontend tests` is the known `UserProfilePage.test.jsx` regression on `main` (PR #780 has the fix). Three-way merge against current `main` is clean — only `ProfileTabScreen.tsx` actually changes. Approving.

PR #778, opened 2026-05-12
feat(mobile): JWT refresh on token_not_valid 401 + graceful logout (closes #405)

### LINKED ISSUE #405: Mobile JWT: handle token_not_valid 401, add refresh + graceful logout

CaganIslam:
## Summary
- `httpClient.ts`: every authenticated request now detects SimpleJWT `401 { code: "token_not_valid" }`, fires a single-flight `POST /api/auth/refresh/` (concurrent 401s share the same promise via `refreshInFlight`), persists the new access token, and replays the original request exactly once.
- On missing refresh token or a 401 from `/api/auth/refresh/` itself, the client clears `token` / `refresh` / `user` from AsyncStorage and throws a typed `AuthExpiredError`. No navigation happens from the http layer.
- `AuthContext.tsx` registers an `authExpiredHandler` on mount via `setAuthExpiredHandler` so an unrecoverable session collapses `user` / `token` to `null`. Existing screens that already guard on `isAuthenticated` redirect to Login on their own.
- Non-401 responses, non-`token_not_valid` 401s, anonymous requests, and `parseJsonResponse` semantics are preserved unchanged.
- AsyncStorage keys touched: read `token` + `refresh`, write `token` on refresh, multiRemove `[token, refresh, user]` on graceful logout.

Closes #405

## Test plan
- [x] `npx tsc --noEmit` in `app/mobile` is clean.
- [x] Metro bundle build: `GET http://localhost:8081/index.bundle?platform=android&dev=true&minify=false` → HTTP 200, bundle contains `AuthExpiredError`.
- [x] Live smoke probe vs `https://genipe.app`: mangled access token → `GET /api/users/me/` returns 401 with `code: "token_not_valid"`; bogus refresh → `POST /api/auth/refresh/` returns 401 → exercises the graceful-logout branch (clears AsyncStorage + fires `authExpiredHandler`).
- [ ] Manual on device: stale token in storage → first authenticated screen load triggers refresh transparently if refresh is still valid; if not, user lands on Login without needing `adb shell pm clear`.
- [ ] Fire several authenticated requests in parallel right after expiry — only one `POST /api/auth/refresh/` should appear in network logs.

### DIFF
--- a/app/mobile/src/context/AuthContext.tsx
+++ b/app/mobile/src/context/AuthContext.tsx
@@ -8,6 +8,7 @@ import React, {
   useState,
 } from 'react';
 import { logoutRequest } from '../services/authService';
+import { setAuthExpiredHandler } from '../services/httpClient';
 import type { AuthUser } from '../services/mockAuthService';
 
 const TOKEN_KEY = 'token';
@@ -71,6 +72,26 @@ export function AuthProvider({ children }: { children: React.ReactNode }) {
     },
     [],
   );
+
+  /**
+   * Lightweight local-only logout used when httpClient detects an expired
+   * session (token_not_valid + refresh failed). The full `logout()` below
+   * also tries to blacklist the refresh server-side, but at this point the
+   * refresh is already invalid so we skip the network call and just collapse
+   * local state. AsyncStorage was already cleared by httpClient.
+   */
+  const localLogout = useCallback(() => {
+    setUser(null);
+    setToken(null);
+  }, []);
+
+  useEffect(() => {
+    // Register with the http layer so a 401 token_not_valid that can't be
+    // refreshed turns into `user === null` in React state. Screens guarded
+    // by `isAuthenticated` will then redirect to Login on their own.
+    setAuthExpiredHandler(() => localLogout());
+    return () => setAuthExpiredHandler(null);
+  }, [localLogout]);
 
   const logout = useCallback(async () => {
     // Best-effort: blacklist the refresh on the backend so a leaked access
--- a/app/mobile/src/services/httpClient.ts
+++ b/app/mobile/src/services/httpClient.ts
@@ -2,6 +2,35 @@ import AsyncStorage from '@react-native-async-storage/async-storage';
 import { API_BASE_URL } from '../config/apiBase';
 
 const TOKEN_KEY = 'token';
+const REFRESH_KEY = 'refresh';
+const USER_KEY = 'user';
+
+/**
+ * Typed error surfaced when the session is unrecoverable — either there is no
+ * refresh token at the moment a `token_not_valid` 401 lands, or the refresh
+ * endpoint itself rejected our refresh token. The http layer intentionally
+ * does NOT navigate; it just clears local auth state and throws this so the
+ * screen / AuthContext can route the user to Login.
+ */
+export class AuthExpiredError extends Error {
+  readonly isAuthExpired = true;
+  constructor(message = 'Session expired') {
+    super(message);
+    this.name = 'AuthExpiredError';
+  }
+}
+
+/**
+ * AuthContext registers a callback here on mount so that when the http layer
+ * decides the session is dead it can notify React state (user → null) without
+ * the http layer itself knowing about navigation or React. Kept as a single
+ * subscriber on purpose: there is only ever one AuthProvider mounted.
+ */
+type AuthExpiredHandler = () => void;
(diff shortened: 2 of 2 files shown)


ErenCanOzkaya (approved):
This is exactly the right shape for mobile JWT refresh — mirrors the web `apiClient` interceptor (single-flight + replay-once) without coupling the HTTP layer to React.

## Highlights

- **Single-flight `refreshInFlight`** prevents the obvious thundering herd: N concurrent authenticated requests all racing a `token_not_valid` 401 share one refresh promise, then each replays its own original request with the fresh access token. Identical to `apiClient.interceptors.response.use(...)` on web.
- **Body-safe detection** — `isTokenNotValid401` reads `res.clone()` so the original `Response` stays consumable by the caller if we end up returning it (non-`token_not_valid` 401s, e.g. permission denied on public endpoints, pass through untouched).
- **`AuthExpiredError`** with `isAuthExpired = true` makes it easy for screens to branch without doing `err.message === '…'` string-matching. Typed, idiomatic.
- **`setAuthExpiredHandler` pub-sub** keeps `httpClient` free of React imports. `AuthContext` subscribes in a `useEffect` and unsubscribes on unmount; single subscriber on purpose.
- **No refresh loop** — replay-once. If the just-minted access token also returns 401, we collapse the session instead of retrying forever. Good defensive choice.
- **`buildHeaders(token)` last-write-wins for caller headers** — preserves the existing API where callers can override `Content-Type` etc.
- **`apiPatchFormData`** correctly routed through the same path (was previously bypassing the auth helpers), and `isJsonBody: false` means we don't accidentally stamp a `Content-Type: application/json` header on multipart bodies.
- **`localLogout` in `AuthContext`** correctly skips the network `logoutRequest` since the refresh token is already known dead at that point.

## Minor

- `AuthExpiredError` is thrown but not caught by every caller. That's intentional (the global handler does the state collapse) but downstream callers that try/catch around `apiGetJson` will see this error too. Worth flagging in the `httpClient` JSDoc that callers can check `err instanceof AuthExpiredError` if they want to suppress their own error toasts. Not blocking — current behaviour is correct, just a doc-nit.
- One subtle invariant: `gracefulLogout` clears `TOKEN_KEY`, `REFRESH_KEY`, `USER_KEY` directly via `AsyncStorage.multiRemove`. If `AuthContext` ever adds a fourth key, the http layer won't clear it. Could live as a single `STORAGE_KEYS` constant somewhere shared — future cleanup, not for this PR.

## CI

Red `Frontend tests` is the known `UserProfilePage.test.jsx` regression on `main` (PR #780 has the one-line fix). Three-way merge clean: only `httpClient.ts` and `AuthContext.tsx` actually change. Approving.

PR #779, opened 2026-05-12
feat(mobile): Map Discovery UX — sized marker badges, animated camera focus, region search pill, a11y (mobile half of #474)

CaganIslam:
## Summary
Mobile half of #474 Map Discovery UX. Region pins are now custom dot markers whose diameter scales with recipe count (24 / 32 / 42px) and that show the count inside; tapping a pin runs a 450ms `animateCamera` zoom-in before navigating to `RegionMapDetail`. Added a top "Search regions →" pill that jumps to the existing `Search` screen pre-focused on region querying, plus better empty/loading copy and accessibility labels on markers and the hint card.

Markers use the custom-View + `renderToHardwareTextureAndroid` + `needsOffscreenAlphaCompositing` + dual `tracksViewChanges` pattern we landed in #769 for the heritage map (per-marker tracking state in a new `RegionDotMarker` component). If round markers ever regress to squares on a device, the fallback is to drop the custom child and go back to `pinColor` strings — call it out in QA.

The web half (`MapPage.jsx`) is a separate scope. `Refs #474` (not Closes).

## Test plan
- [ ] Open Discover by region tab on Android — pins render as round mustard circles with the recipe count inside, not squares
- [ ] Pin sizes look distinct: 1-recipe regions are visibly smaller than 6+ regions
- [ ] Tap a pin — map smooth-zooms onto it (~450ms) and then `RegionMapDetail` opens
- [ ] Focused pin (after tap, before navigation completes) turns green
- [ ] Tap the top "Search regions →" pill — navigates to the Search screen
- [ ] With zero pins, hint card shows the "still seeding coordinates" copy
- [ ] TalkBack/VoiceOver on a marker reads "<region name>, N recipes"
- [ ] iOS: same round-marker behaviour, no regressions in zoom controls or the bottom Ingredient routes pill

### DIFF
--- a/app/mobile/src/components/map/RegionDotMarker.tsx
+++ b/app/mobile/src/components/map/RegionDotMarker.tsx
@@ -1,0 +1,120 @@
+import { useEffect, useState } from 'react';
+import { StyleSheet, Text, View } from 'react-native';
+import { Marker } from 'react-native-maps';
+import { tokens } from '../../theme';
+import type { RegionPin } from '../../services/mapDataService';
+
+/** Three-tier size scale for region pins by recipe count. */
+export function markerSizeForRecipeCount(count: number): number {
+  if (count >= 6) return 42;
+  if (count >= 2) return 32;
+  return 24;
+}
+
+/** Just the colored circle with the count number inside.
+ *
+ * Wrapped with `renderToHardwareTextureAndroid` + `needsOffscreenAlphaCompositing`
+ * so Android keeps the rounded corners when it snapshots the custom marker view
+ * into its bitmap. Without this, borderRadius gets dropped and the marker shows
+ * up as a square (same fix as #769 on the heritage map). */
+function RegionDot({
+  size,
+  isFocused,
+  count,
+}: {
+  size: number;
+  isFocused: boolean;
+  count: number;
+}) {
+  const fontSize = size >= 42 ? 14 : size >= 32 ? 12 : 10;
+  return (
+    <View
+      collapsable={false}
+      renderToHardwareTextureAndroid
+      needsOffscreenAlphaCompositing
+      style={[styles.outer, { width: size + 8, height: size + 8 }]}
+    >
+      <View
+        style={[
+          styles.dot,
+          {
+            width: size,
+            height: size,
+            borderRadius: size / 2,
+            backgroundColor: isFocused
+              ? tokens.colors.accentGreen
+              : tokens.colors.accentMustard,
+          },
+        ]}
+      >
+        <Text
+          allowFontScaling={false}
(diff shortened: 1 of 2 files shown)


ErenCanOzkaya (approved):
Map polish hangs together well — sized markers + count overlay + animated camera + search pill all read as one coherent UX rather than four separate fixes.

## Highlights

- `markerSizeForRecipeCount(count)` three-tier scale (24/32/42) gives quick visual rank without overcrowding the map. The conditional `fontSize` inside the dot (10/12/14) is a thoughtful touch — large pins get a readable count, small ones don't get a giant glyph crammed in.
- `animateCamera({ duration: 450 })` runs in parallel with `navigation.navigate(...)`; the comment makes the parallel-fire intentional (continuous transition, no jank). `FOCUS_ZOOM = 5` documented with the Google Maps zoom convention — good.
- `RegionDotMarker` is a clean lift — single source of truth for sized + focused-aware region pins, reusable if we ever wallpaper this style onto a different map.
- `nestedScrollEnabled` + better loading copy + a11y labels on the search pill are small wins that compound.

## Worth confirming — Android marker square clipping

The custom round-view path (`borderRadius` inside a Marker) is the exact pattern that PR #769 explicitly moved away from on `HeritageMapScreen` after `tracksViewChanges` + `renderToHardwareTextureAndroid` + `needsOffscreenAlphaCompositing` weren't enough — bitmap snapshot dropped the rounded corners and rendered the marker as a square. The fix there was to switch to the platform `pinColor` API.

Here you're using the same hardware-texture + offscreen-alpha combo (and the comment even cites #769), but without `tracksViewChanges` toggling. Since this PR's commit is from after #769's merge fix, I'm assuming you've verified the count-overlay variant survives the snapshot on Android — `pinColor` can't carry text inside the pin so platform default isn't an option here.

Couple of things I'd double-check on an Android device before merging:
- Initial render with all pins on screen: do any render as square?
- After zoom-in via `animateCamera`, do the resized pins stay round?
- If they're square: the fallback I'd reach for is a `tracksViewChanges` toggle (start true → flip false after a short timeout) like the earlier heritage version, or a single text label outside the dot. Both keep the count visible.

If Android renders cleanly, ignore the above.

## Minor

- `setFocused(pin)` runs immediately, then `animateCamera`, then `navigate`. On a fast device the focus highlight on the previous pin might be visible for a frame mid-animation before navigation pops the screen. Not a real bug, just a curiosity — could clear focus on screen blur in a follow-up if it ever shows up in QA.

## CI

Red `Frontend tests` is the known `UserProfilePage.test.jsx` regression on `main` (PR #780 has the fix). Three-way merge clean: only `RegionDotMarker.tsx` (new) and `MapDiscoveryScreen.tsx` actually change. Approving — confirm the Android marker question once you've eyeballed a device build.

PR #781, opened 2026-05-12
feat(mobile): passport screen scaffolding — cover, stats, tab pills, placeholders (closes #598)

CaganIslam:
## Summary
- New `PassportScreen` (mobile) scaffolds the passport profile experience: cover band, identity header, 4-stat bar, 5-tab pill row with placeholder bodies.
- New `passportService.fetchPassport(username)` hits `GET /api/users/<username>/passport/` and normalizes missing keys to empty arrays / null / 0.
- Entry point added: small mustard "🛂 Passport" pill on the existing `UserProfileScreen` identity card, passing `username` + `isOwn` to the new route.
- Sibling PRs (#600 cover, #601 stamps, #603 cultures, #604 timeline, #605 quests) replace the placeholder bodies in follow-up work.

## Test plan
- [ ] Open a user's profile screen, see the 🛂 Passport pill on the right of the avatar row
- [ ] Tap the pill, Passport screen loads (Level/points pill, theme line, identity card)
- [ ] Verify all 5 tab pills (Stamps / Cultures / Map / Timeline / Quests) toggle and show the placeholder copy + counts
- [ ] Stats bar shows 4 numeric stats from the API response (recipes_tried, stories_saved, cultures_count, heritage_shared)
- [ ] Pull-to-rebuild / re-focus screen re-fetches the passport

Closes #598

PR #759, opened 2026-05-12
feat(mobile): star rating UI on recipe detail and cards + notification icon (closes #735, closes #738)

CaganIslam:
## Summary
- Adds a reusable `StarRatingRow` component (5-star, mustard fill, sm/lg sizes, read-only mode, tap-same-star-to-clear) used on Recipe Detail and recipe Search cards.
- Recipe Detail: interactive star row with optimistic update against `POST /api/recipes/:id/rate/` and `DELETE` on clear. Own recipes render read-only; unauthenticated taps route to Login. Caption shows `4.2 ★ · N ratings` or `Not rated yet`.
- Search result cards (recipe kind) render a small read-only star row when `average_rating` is present in the response; hidden otherwise so we never fabricate.
- New `ratingService.ts` with defensive `toNum` for the DRF DecimalField (string `"4.20"`) shape.

## Notes
- **#738 NOT addressed in this PR:** the mobile app currently has no notifications screen, list component, or service. Filed **#760** to build the scaffolding first. #738 stays open and depends on #760.
- **Cards skipped:** `RecommendationsRail` (home recommendations) and `searchService` story results do not expose `average_rating` from the backend response — left untouched rather than fabricated.
- **Overlaps with #755 (bookmark cluster)** on `RecipeDetailScreen.tsx` and `SearchResultCard.tsx` — merge-time conflict expected if #755 lands first. Concerns are separate (save pill vs star row), so the merge will be mechanical.

## Test plan
- [ ] Open a recipe you did NOT author while signed in — tap a star, average updates optimistically, caption shows new count.
- [ ] Tap the same star again — rating clears (DELETE) and detail refetches.
- [ ] Open your own recipe — stars render but are not tappable.
- [ ] Sign out, open any recipe — stars render read-only; tapping any star routes to Login.
- [ ] Search for recipes — when backend includes `average_rating`, a small star row appears on the card.

Closes #735

PR #783, opened 2026-05-12
feat(mobile): "I Tried This" / "Add to Passport" / "Save to Passport" action buttons (closes #599)

CaganIslam:
## Summary
Adds three passport-action pills to mobile detail screens (#599). Recipe detail gains an "I Tried This" pill (green active state, fork/check icon) and an "Add to Passport" pill (mustard active state, passport icon) next to the existing Save bookmark. Story detail gains a single "Save to Passport" pill (mustard active state). Each pill is auth-gated (taps from anonymous users route to Login), uses optimistic toggle with rollback + inline error toast, and reads its active state from the recipe/story payload (`is_tried`, `in_passport`, `saved_to_passport`) so a fresh load reflects existing stamps.

A new service `passportActionService.ts` wraps the three endpoints. Types were extended with the optional flags and a Jest test covers `tryRecipe`, `addRecipeToPassport`, and `saveStoryToPassport` (right path + response parsing + empty-body fallback to the optimistic value).

## Endpoints used
The brief listed several candidate paths and asked us to probe. Against `https://genipe.app` on 2026-05-12 every candidate returned 404 (backend `#584` has not been deployed to the shared host yet — verified with a freshly-registered user; provided `ayse@example.com` credentials also failed with "Invalid credentials"). The mobile client targets the canonical paths from the brief so it goes live the moment backend lands:

- `POST /api/recipes/:id/i-tried-this/` — toggles `is_tried`
- `POST /api/recipes/:id/add-to-passport/` — toggles `in_passport`
- `POST /api/stories/:id/save-to-passport/` — toggles `saved` / `saved_to_passport` (service accepts either field name in the response)

If backend chose different route names, only the path strings in `passportActionService.ts` need to change. The service is defensive: if the body is empty the UI falls back to the optimistic value rather than flickering.

## Active state sourcing
Active state is read from the recipe/story payload (`recipe.is_tried`, `recipe.in_passport`, `story.saved_to_passport`). Backend #584 is expected to surface these fields on the detail responses, mirroring how `is_bookmarked` is surfaced today. No extra round-trip on mount.

## Test plan
 [ ] `cd app/mobile && npx tsc --noEmit` is clean
 [ ] `cd app/mobile && npx jest __tests__/services/passportActionService.test.ts` — 4 tests pass
 [ ] Metro `index.bundle?platform=ios&dev=false&minify=true` returns HTTP 200
 [ ] Anonymous user opens a recipe — only the Save bookmark pill is visible, no Tried/Passport pills
 [ ] Signed-in user opens a recipe — both pills appear; tapping "I Tried This" flips to green "Tried" with a check; tapping again flips back
 [ ] Same flow for "Add to Passport" → mustard "In Passport"
 [ ] Force a backend error (offline / fake 500) — pill rolls back and a toast shows
 [ ] Open a story signed-out — no Save-to-Passport pill; signed-in shows the pill and toggles with mustard active state

Closes #599

### DIFF
--- a/app/mobile/__tests__/services/passportActionService.test.ts
+++ b/app/mobile/__tests__/services/passportActionService.test.ts
@@ -1,0 +1,51 @@
+import {
+  saveStoryToPassport,
+  tryRecipe,
+} from '../../src/services/passportActionService';
+import { apiPostJson } from '../../src/services/httpClient';
+
+jest.mock('../../src/services/httpClient', () => ({
+  apiPostJson: jest.fn(),
+}));
+
+const mockedPost = apiPostJson as jest.MockedFunction<typeof apiPostJson>;
+
+describe('tryRecipe', () => {
+  beforeEach(() => mockedPost.mockReset());
+
+  it('POSTs to the passport try endpoint and returns the parsed flag', async () => {
+    mockedPost.mockResolvedValueOnce({ is_tried: true });
+    const out = await tryRecipe(1, true);
+    expect(mockedPost).toHaveBeenCalledWith('/api/passport/recipes/1/try/', {});
+    expect(out).toEqual({ is_tried: true });
+  });
+
+  it('falls back to the optimistic value when the backend returns no body', async () => {
+    mockedPost.mockResolvedValueOnce(null as unknown as { is_tried: boolean });
+    const out = await tryRecipe('1', false);
+    expect(out).toEqual({ is_tried: false });
+  });
+});
+
+describe('saveStoryToPassport', () => {
+  beforeEach(() => mockedPost.mockReset());
+
+  it('POSTs to the passport save endpoint and accepts either response field', async () => {
+    mockedPost.mockResolvedValueOnce({ saved_to_passport: true });
+    const out = await saveStoryToPassport(42, true);
+    expect(mockedPost).toHaveBeenCalledWith('/api/passport/stories/42/save/', {});
+    expect(out).toEqual({ saved: true });
+  });
+
+  it('also accepts a `saved` field on the response', async () => {
+    mockedPost.mockResolvedValueOnce({ saved: true });
+    const out = await saveStoryToPassport(42, false);
+    expect(out).toEqual({ saved: true });
+  });
+
+  it('falls back to the optimistic value when the backend returns no body', async () => {
+    mockedPost.mockResolvedValueOnce(null as unknown as { saved: boolean });
+    const out = await saveStoryToPassport('99', true);
+    expect(out).toEqual({ saved: true });
+  });
+});
--- a/app/mobile/src/screens/RecipeDetailScreen.tsx
+++ b/app/mobile/src/screens/RecipeDetailScreen.tsx
@@ -20,6 +20,7 @@ import { fetchCulturalFactsByRegion, type CulturalFact } from '../services/cultu
 import type { RootStackParamList } from '../navigation/types';
 import { fetchCheckedIngredients, toggleCheckedIngredient } from '../services/checkOffService';
 import { toggleBookmark } from '../services/bookmarkService';
+import { tryRecipe } from '../services/passportActionService';
 import { fetchRecipeById } from '../services/recipeService';
 import { fetchStoriesForRecipe, type StoryListItem } from '../services/storyService';
 import { fetchConversion } from '../services/unitConversionService';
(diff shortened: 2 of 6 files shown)


ErenCanOzkaya (changes requested):
The pill UX, auth gating, and optimistic-state plumbing all read well. Two things to address before this can land:

## Blocking #1 — endpoint paths don't match the merged backend

`passportActionService.ts` POSTs to three paths that do not exist in the deployed backend. Verified by grepping `app/backend/apps/passport/urls.py` on `main`:

```python
path('passport/recipes/<str:recipe_id>/try/', TryRecipeView.as_view(), ...)
path('passport/stories/<str:story_id>/save/', SaveStoryView.as_view(), ...)
```

| This PR | Real route |
|---|---|
| `POST /api/recipes/:id/i-tried-this/` | `POST /api/passport/recipes/:id/try/` |
| `POST /api/recipes/:id/add-to-passport/` | — does not exist — |
| `POST /api/stories/:id/save-to-passport/` | `POST /api/passport/stories/:id/save/` |

The "Add to Passport" affordance has no corresponding backend route: the backend deliberately collapses "tried" and "saved to passport" into the single `/try/` action — `TryRecipeView` updates the stamp, advances quests, and writes the timeline event in one shot (see `apps/passport/services.py#record_recipe_try`). The same call has already been made on the web side (PR #764) where `addRecipeToPassport` ended up being removed for the same reason.

Two ways forward:
- **(a) preferred:** drop the `Add to Passport` pill on `RecipeDetailScreen` and ship `I Tried This` + `Save to Passport` (story). Matches the actual backend and matches the web app's resolution.
- **(b)** keep both pills visually but have them call the same `/try/` endpoint. Less honest UX-wise; we agreed on (a) when web went through this.

The remaining two paths need updating to the real ones:
```ts
`/api/passport/recipes/${recipeId}/try/`
`/api/passport/stories/${storyId}/save/`
```

Response shape note: real responses are richer than just `{ is_tried }` / `{ saved }` — `TryRecipeView` returns the full passport bundle plus `leveled_up`, `affected_stamps`, `newly_completed_quests`. For the pill UX you only need to know "did the action succeed" — `coerceBool(data?.is_tried, nextValue)` could be simplified to "200 ⇒ optimistic value wins", since the endpoint is idempotent and the truthful current state lives in the passport payload (already fetched by `fetchPassport` in #781).

## Blocking #2 — conflict against current main

`RecipeDetailScreen.tsx` conflicts with the star-rating block that landed in #759. Lines 497-622: HEAD adds an `actionRow` containing bookmark + tried + passport pills, `origin/main` has a `ratingBlock` followed by a standalone bookmark pill.

Resolution should keep both, with bookmark consolidated into the new `actionRow` (to avoid showing the bookmark twice):

```jsx
<View style={styles.ratingBlock}>
  <StarRatingRow ... />
  <Text style={styles.ratingCaption}>{...}</Text>
</View>

<View style={styles.actionRow}>
  <Pressable /* bookmark */ ... />
  {isAuthenticated && (
    <>
      <Pressable /* I Tried This */ ... />
      {/* drop the Add to Passport pill per blocker 1(a) */}
    </>
  )}
</View>
```

The styles conflict at line 933 is straightforward — keep both `actionRow` and the new `ratingBlock` + `ratingCaption`:

```ts
actionRow: { marginTop: 12, flexDirection: 'row', flexWrap: 'wrap', alignItems: 'center', gap: 8 },
ratingBlock: { marginTop: 14, gap: 4 },
ratingCaption: { fontSize: 13, color: tokens.colors.textMuted, fontWeight: '600' },
```

## Highlights once the above are sorted

- Optimistic pill state with `nextValue` fallback in `coerceBool` is the right idea — the pill settles into a coherent state even if the response body is empty.
- Auth gating (`isAuthenticated ? <pill> : null`) is consistent with how the navbar / chat tray handle anonymous users.
- Disabled state during the in-flight POST (`disabled={triedBusy}`) + accessibility states (`busy`, `selected`) are nicely covered.
- `Story` type gets the optional `saved_to_passport` flag — consistent with how `is_tried` is handled on recipes.

Happy to re-review once endpoints are corrected and the rebase lands.

ErenCanOzkaya (approved):
Re-approving on top of the rebase + endpoint fixes I pushed. Tests are green locally (`npx jest`: 51/51, `npx tsc --noEmit`: clean). Both blockers from the previous review are addressed.

mustafaocakxyz (approved):
LGTM nice implementation

PR #784, opened 2026-05-12
feat(mobile): culture grid + detail screen for passport cultures (closes #602)

CaganIslam:
## Summary
- Adds `passportCultureService` with `fetchCultureDetail(username, cultureName)`; normalizes the backend `culture_summaries` payload to the canonical `CultureSummary` shape and uses `toNum` for defensive coercion.
- Ships `CultureGrid` (2-column `FlatList`) with a rarity colour swatch per cell (bronze/silver/gold/emerald/legendary) and an empty-state copy for users with no cultures yet.
- Ships `CultureDetailScreen` with display-font header, rarity badge, and a four-block stats grid (recipes_tried / stories_saved / ingredients_discovered / heritage_recipes); `LoadingView`, `ErrorView` with retry, and a "not found" empty state.
- Wires `CultureDetail` into `RootStackParamList` + `PublicStackNavigator` so the grid can push the detail route from anywhere passport #598 mounts it.
- Tests: service finds known culture, returns null on miss, normalizes mixed field names, drops malformed rows; grid renders cells, shows empty state, fires `navigation.navigate` on tap.

## culture_summaries shape (discovered)
Probed `apps/passport/services.py#culture_summaries` plus the frontend mock — backend currently returns:
```json
{ "culture": "Ottoman", "recipes_tried": 8, "stories_saved": 3, "interactions": 12, "rarity": "gold" }
```
Note: backend uses `culture` / `rarity`; the issue spec uses `culture_name` / `stamp_rarity` and adds `ingredients_discovered` / `heritage_recipes`. The service maps both ways so callers can rely on the spec shape, defaulting missing counters to `0`.

Could not log in as `ayse@example.com / StrongPassword123!` against `https://genipe.app` (`Invalid credentials`); shape was confirmed from the backend code path and the matching web mock in `app/frontend/src/services/passportService.js`.

## Test plan
- [ ] `cd app/mobile && npx tsc --noEmit` clean
- [ ] `cd app/mobile && npx jest __tests__/services/passportCultureService __tests__/components/CultureGrid` — 8 tests pass
- [ ] Metro bundle (`/index.bundle?platform=ios`) returns HTTP 200
- [ ] From a screen that has cultures, tap a cell → pushes `CultureDetail` with the right params; stats and rarity badge render

Closes #602

### DIFF
--- a/app/mobile/__tests__/components/CultureGrid.test.tsx
+++ b/app/mobile/__tests__/components/CultureGrid.test.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react-native';
+
+const mockNavigate = jest.fn();
+jest.mock('@react-navigation/native', () => ({
+  useNavigation: () => ({ navigate: mockNavigate }),
+}));
+
+import { CultureGrid } from '../../src/components/passport/CultureGrid';
+import type { CultureSummary } from '../../src/services/passportCultureService';
+
+const makeCulture = (over: Partial<CultureSummary> = {}): CultureSummary => ({
+  culture_name: 'Ottoman',
+  stamp_rarity: 'gold',
+  recipes_tried: 8,
+  stories_saved: 3,
+  ingredients_discovered: 0,
+  heritage_recipes: 0,
+  ...over,
+});
+
+describe('CultureGrid', () => {
+  beforeEach(() => mockNavigate.mockReset());
+
+  it('renders the empty-state copy when no cultures are provided', () => {
+    const { getByText, queryByText } = render(<CultureGrid cultures={[]} username="ayse" />);
+    expect(
+      getByText('No cultures discovered yet. Try recipes from new regions to unlock cultures.'),
+    ).toBeTruthy();
+    expect(queryByText('Ottoman')).toBeNull();
+  });
+
+  it('renders one cell per culture', () => {
+    const cultures = [
+      makeCulture({ culture_name: 'Ottoman' }),
+      makeCulture({ culture_name: 'Aegean', stamp_rarity: 'silver', recipes_tried: 5 }),
+      makeCulture({ culture_name: 'Mediterranean', stamp_rarity: 'bronze', recipes_tried: 3 }),
+    ];
+    const { getByText } = render(<CultureGrid cultures={cultures} username="ayse" />);
+    expect(getByText('Ottoman')).toBeTruthy();
+    expect(getByText('Aegean')).toBeTruthy();
+    expect(getByText('Mediterranean')).toBeTruthy();
+  });
+
+  it('navigates to CultureDetail when a cell is tapped', () => {
+    const cultures = [makeCulture({ culture_name: 'Ottoman' })];
+    const { getByLabelText } = render(<CultureGrid cultures={cultures} username="ayse" />);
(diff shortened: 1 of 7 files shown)


ErenCanOzkaya (commented):
Strong work — the service layer correctly hits the real backend endpoint (`GET /api/users/<username>/passport/`) and normalises the `culture_summaries` payload to the issue contract (`culture_name` / `stamp_rarity`) while accepting the backend's actual field names (`culture` / `rarity`) as fallbacks. Good defensive shape: `ingredients_discovered` and `heritage_recipes` aren't on the backend yet, so they correctly default to `0`. This is the exact contract-matching that PR #783 should also be doing.

`CultureGrid` 2-column FlatList with rarity-coloured swatch and empty-state copy is clean; the same `RARITY_COLORS` map gets reused by `CultureDetailScreen` (importing from `CultureGrid`), which is fine for this scope — if a third consumer shows up, it's worth hoisting to a `utils/rarityColors.ts`.

`CultureDetailScreen` cancellation pattern (`let cancelled = false; return () => { cancelled = true }`) is correct in every branch (.then / .catch / .finally) — no unmounted setState risk.

`fetchCultureDetail` looking up by lowercased + trimmed name is the right shape for URL-passed identifiers; matches how the web side ended up keying cultures.

## Blocking — trivial merge conflicts

Two files conflict against current `main`, both because #776 (Notifications scaffolding) and this branch independently added a new screen to the same lists:

**`app/mobile/src/navigation/types.ts`** — keep both:
```ts
RegionMapDetail: { regionId: number; regionName: string };
IngredientMigrationMap: undefined;
CultureDetail: { username: string; cultureName: string };
Notifications: undefined;
```

**`app/mobile/src/navigation/PublicStackNavigator.tsx`** — keep both `<Stack.Screen>` entries. Order doesn't matter, both can follow the existing `IngredientMigrationMap` block.

These are the kind of conflicts `git rebase origin/main` resolves in a single keystroke (accept both sides). Once rebased, CI should also pick up green status (the workflow hasn't run yet on this PR).

## Minor

- The `username` prop on `CultureGrid` is forwarded for navigation parameters (`navigation.navigate('CultureDetail', { username, cultureName })`). If a future entry point ever renders the grid without a known username (e.g. on a logged-out preview), the navigation will crash on missing param. Not a problem today — `PassportScreen` always knows the username — just worth a `username?:` + a `if (!username) return` guard later.
- `fetchCultureDetail` fires a full passport fetch for a single culture lookup. Cheap right now (backend bundles everything in one call), but if cultures grow we may want a dedicated detail endpoint. Out of scope for this PR.

## Highlights

- The grid is 2-column with `scrollEnabled={false}` so it nests cleanly inside `PassportScreen`'s parent ScrollView — correct.
- Empty state copy guides the user toward the action that unlocks the next culture — nice product touch.
- Rarity swatch on every card gives the grid a visual rank at a glance.
- Detail screen lays out all five rollup fields (recipes tried, stories saved, ingredients, heritage, current rarity) cleanly.

Happy to re-review once the navigation rebase lands.

ErenCanOzkaya (approved):
Approving on top of the rebase I pushed — the trivial keep-both conflict resolution on `types.ts` and `PublicStackNavigator.tsx` is the only change versus your branch tip; the rest of the diff (service, components, screen, tests) is untouched and reads cleanly.

mustafaocakxyz (approved):
Great job both, thanks

PR #785, opened 2026-05-12
fix(backend): add primary-key tiebreaker to recipe and story list ordering (#770)

laternis:
## Summary
- `apps/stories/models.py`, `apps/recipes/models.py`: `seed_canonical` inserts every recipe/story inside one transaction, so many rows share an identical `created_at` to the microsecond. Without a deterministic tiebreaker the database may return rows in a different order between paginated queries, so a row near a page boundary can appear on two consecutive pages. `Story.Meta.ordering` and `Recipe.Meta.ordering` now end with `-id` (`['-created_at', '-id']`).
- `apps/recipes/views.py`: `RecipeViewSet`'s `Count('story_links'...)` / `Count('bookmarks', distinct=True)` annotations add a `GROUP BY`, and Django does not apply a model's default ordering to `GROUP BY` queries — so `Meta.ordering` alone did not fix `/api/recipes/`. `RecipeViewSet.get_queryset` now re-applies `.order_by('-created_at', '-id')` explicitly. The `regions/<id>/recipes/` action gets the same `-id` tiebreaker.
- Migrations: `recipes/0029_list_ordering_pk_tiebreaker_770.py`, `stories/0013_list_ordering_pk_tiebreaker_770.py` — `AlterModelOptions` only, no schema change.
- Tests: `apps/stories/tests_pagination.py`, `apps/recipes/tests_pagination.py` — assert the `Meta.ordering` contract, walk `/api/stories/` and `/api/recipes/` page by page asserting each id appears exactly once, and assert the paginated order is deterministic (`-created_at, -id`).

## Test plan
- [x] `python manage.py test apps.stories.tests_pagination apps.recipes.tests_pagination` — 6 tests, OK (verified failing before the fix, passing after)
- [x] `python manage.py test` — 874 tests, OK (no regressions)
- [ ] After `seed_canonical`, walk `/api/stories/` and `/api/recipes/` page by page — no id appears on two consecutive pages

## Notes
- Migration numbers (`recipes/0029`, `stories/0013`) may collide if another backend branch merges first; rebase + bump if so.
- `Comment` / `Notification` and other list-style models were left untouched — the acceptance criteria scope this to `Story` and `Recipe`. Easy follow-up if wanted.

Closes #770.


roboticrustacean (approved):
lgtm

Issue #365, opened 2026-04-18
M6-12 — Signed APK build


ErenCanOzkaya:
## Description
Reproducible signed APK produced by CI, ready for manual installation on reviewer devices.

## Deliverables
- CI signing pipeline
- Signed APK artifact attached to release

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

## References
- Draft requirement D.10.1

## Notes

PR #792, opened 2026-05-12
chore(devops): finish CI coverage reporting (mobile job, html artifact, config, docs) (#534)

laternis:
## Summary
- tests.yml: added a mobile-tests job (Node 22, jest --coverage) that prints a coverage table to the PR summary and uploads app/mobile/coverage as an artifact.
- tests.yml: backend job now runs coverage html and uploads app/backend/htmlcov as an artifact.
- package.json (frontend + mobile): added jest.collectCoverageFrom so coverage only counts app code under src/, excluding mocks and test files.
- docs/coverage-reporting.md: documented how to run coverage per layer, where CI artifacts land, and a coverage breakdown table. The GitHub wiki page should be created from this doc.

## Test plan
- [x] tests.yml is valid YAML
- [x] Backend / frontend / mobile jobs pass in CI
- [x] Coverage tables appear in the PR job summary for all three layers
- [x] htmlcov and mobile coverage artifacts are downloadable from the run

## Notes
- Backend .coveragerc was already correct and is unchanged.
- mobile-apk.yml is owned by a sibling PR (#365) and is untouched here.

Closes #534.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,10 +48,19 @@ jobs:
         if: always()
         run: |
           coverage combine || true
+          coverage html --omit='*/migrations/*,*/tests*' --skip-empty || true
           echo "## Backend coverage" >> "$GITHUB_STEP_SUMMARY"
           echo '```' >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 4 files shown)


roboticrustacean (approved):
lgtm

PR #793, opened 2026-05-12
feat(frontend): region-centric pins, top-region center, region panel (#668)

Daglar1500:
## Summary

heritageMidpoint.js: Replaced `computeHeritageMidpoint` (geometric centroid of all members) with `groupByRegion` and `topRegion`. `groupByRegion` collapses locatable members into per-region buckets with averaged coordinates. `topRegion` picks the region with the most recipes as the semantic center, first-occurrence wins on ties.

HeritageMapPage: Replaced per-member CircleMarkers with per-region pins. Map centers on the top region's coordinates. Polylines fan out from that center to each other region. Clicking any pin (including the center 🏛 marker) opens an in-page region panel that lists the heritage's recipes and stories tagged to that region, with row links to `/recipes/:id` and `/stories/:id`.

HeritageMapPage.css: Added split-grid layout (map left, 320 px panel right, stacks on mobile ≤ 700 px) and panel component styles consistent with the existing MapPage panel.

Tests: Rewrote `heritageMidpoint.test.js` (13 cases covering `groupByRegion` and `topRegion`) and `HeritageMapPage.test.jsx` (9 cases covering center position, pin/polyline counts, single-region and zero-region edge cases, and panel open/content on click).

## Test plan
- [ ] `/heritage/:id/map` — map loads, center 🏛 marker sits on the most-recipe region
- [ ] Other regions each have one CircleMarker (no per-member stacking)
- [ ] Polylines connect center → each other region
- [ ] Click center marker → panel opens showing that region's recipes/stories
- [ ] Click a region pin → panel switches to that region's content
- [ ] Row link in panel routes to `/recipes/:id` or `/stories/:id`
- [ ] Heritage with one region only → no polylines, panel still works
- [ ] Heritage with no plottable coordinates → empty-state message, no map
- [ ] `CI=true npm test` → 599/599 pass

Closes #668

### DIFF
--- a/app/frontend/src/__tests__/HeritageMapPage.test.jsx
+++ b/app/frontend/src/__tests__/HeritageMapPage.test.jsx
@@ -1,4 +1,4 @@
-import { render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor, fireEvent } from '@testing-library/react';
 import { MemoryRouter, Route, Routes } from 'react-router-dom';
 import HeritageMapPage from '../pages/HeritageMapPage';
 import * as heritageService from '../services/heritageService';
@@ -12,7 +12,7 @@ jest.mock('react-leaflet', () => ({
   TileLayer: () => null,
   CircleMarker: ({ children, center, eventHandlers }) => (
     <div
-      data-testid="circle-marker"
+      data-testid="region-pin"
       data-center={JSON.stringify(center)}
       onClick={eventHandlers?.click}
     >{children}</div>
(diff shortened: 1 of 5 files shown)


ErenCanOzkaya (approved):
This is the right call for the Heritage Map UX. The original member-midpoint design we shipped in #717 was the simplest thing that could work — visually it tended to land in the ocean or in an empty region, and the per-member pin stacking made multi-recipe regions look noisy. #668's "centre = strongest region" framing is the semantic answer.

## Highlights

- `groupByRegion` correctly groups by region name (`?? ''` covers the "no region" edge case as an empty-key bucket — defensive, plus the empty-region members still get plotted at their own averaged coords rather than being silently dropped).
- `topRegion` reduce with `regionGroups[0]` seed naturally gives first-occurrence-wins tiebreak, matching the issue spec. Insertion order is preserved by `Map` iteration, and the test `preserves insertion order for tie-break` pins that behaviour.
- `locatableMembers` is kept as a separate public export — useful elsewhere and the previously-shipping consumers don't break.
- `computeHeritageMidpoint` is fully removed from the codebase (verified by `git grep` — no stale callers).
- The split-grid CSS layout (`map left, 320px panel right, stacks ≤ 700px`) mirrors `MapPage.css` conventions; visual consistency with the existing region detail flow is exactly what #668 asks for.

## Tests

- `heritageMidpoint.test.js`: 13 cases covering happy path (one group per region), averaging, null-coord drop, empty input, tie-break order, recipe-vs-story tiebreak weighting. Solid coverage.
- `HeritageMapPage.test.jsx`: 9 cases covering map centre, pin/polyline counts, single-region (no polylines), zero-region (empty state), and panel open/content behaviour on pin click. Locally `npm test -- --testPathPattern="Heritage|heritageMid"` returns 42/42 across the heritage suite.

## Minor

- `m.region ?? ''` groups everyone with no region into the same bucket and averages their unrelated lat/lng — could land that bucket anywhere. In practice the backend joins `region` from `Region.name` so an empty region implies an empty/missing `latitude/longitude` too, which `locatableMembers` already filters out before grouping. So the empty-region bucket is unreachable in real data. Not blocking; just worth a comment in the util if it ever shows up in QA.
- Polyline color matches the existing dashed CSS variant — good consistency.
- The center 🏛 emoji + `divIcon` carry-over from the old implementation is the right choice — keeps continuity with the heritage page's iconography.

## CI

Green. Three-way merge against current `main` is clean — only the 5 listed frontend files actually change. Approving.

PR #798, opened 2026-05-12
fix(frontend): null username crash, video validation, submitting guard, orphan recipe (#693, #696, #703, #704)

Daglar1500:
## Summary

- **InboxPage, ThreadPage**: `username[0]` calls crash on null/empty usernames — replaced with `username?.[0]?.toUpperCase() ?? '?'` in both files. Navbar was already guarded.
- **RecipeEditPage**: Video validation checked only the newly selected file (`!video`), ignoring an already-uploaded server video (`recipe.video`). Fix adds `!recipe?.video` to the condition so existing videos satisfy the rule.
- **RecipeCreatePage**: On successful JSON POST but failed media PATCH, the page showed a generic error and left an orphan recipe. Now navigates to the new recipe immediately with a distinct toast ("published but media upload failed"), preventing duplicate submissions on retry.
- **Login, Register, StoryCreate, RecipeCreate**: No submitting guard existed — rapid clicks fired concurrent requests. Added `submitting` boolean state; buttons are disabled and handler is gated while in-flight.

## Test plan

 [ ] Open Inbox as a user whose conversation partner has a null/empty username — no white screen, `?` initial shows
 [ ] Edit a recipe that already has a video, clear the description, save — validation passes
 [ ] Edit a recipe with neither a remote video nor a description — validation still rejects
 [ ] On Recipe Create, spam-click Publish — only one POST in the network log, button goes grey
 [ ] On Login, spam-click Log In — only one auth request fires
 [ ] On Register, spam-click Register — only one registration request fires
 [ ] On Story Create, spam-click Publish — only one POST fires

### DIFF
--- a/app/frontend/src/pages/InboxPage.jsx
+++ b/app/frontend/src/pages/InboxPage.jsx
@@ -117,7 +117,7 @@ export default function InboxPage() {
             <li key={thread.id}>
               <Link to={`/inbox/${thread.id}`} className="thread-row">
                 <div className="thread-avatar" aria-hidden="true">
-                  {thread.otherUser.username[0].toUpperCase()}
+                  {thread.otherUser.username?.[0]?.toUpperCase() ?? '?'}
                 </div>
                 <div className="thread-info">
                   <div className="thread-header-row">
--- a/app/frontend/src/pages/LoginPage.jsx
+++ b/app/frontend/src/pages/LoginPage.jsx
@@ -13,6 +13,7 @@ export default function LoginPage() {
   const [password, setPassword] = useState('');
   const [errors, setErrors] = useState({});
   const [apiError, setApiError] = useState('');
+  const [submitting, setSubmitting] = useState(false);
 
   const animationClass = location.state?.from === 'register' ? 'auth-enter-left' : 'auth-enter-right';
 
@@ -25,6 +26,7 @@ export default function LoginPage() {
 
   async function handleSubmit(e) {
     e.preventDefault();
+    if (submitting) return;
     const errs = validate();
     if (Object.keys(errs).length > 0) {
       setErrors(errs);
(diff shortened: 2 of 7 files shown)


ErenCanOzkaya (approved):
Tight, defensive bug sweep — all four fixes hit real pain points users would see in production. Walking through each:

## #693 — Null username crash on Inbox / Thread

`username?.[0]?.toUpperCase() ?? '?'` is the right shape; matches the guarded pattern Navbar already uses (consistent across the app). Affects both `InboxPage` (thread list avatars) and `ThreadPage` (per-message bubble avatars).

## #696 — Recipe edit video validation

`!description.trim() && !video && !recipe?.video` — the previous condition forced uploading a *new* video on every edit even when the recipe already had one on the server. Adding the `recipe?.video` clause lets the existing video satisfy the rule. Optional chaining covers the load-error case where `recipe` is still null. Correct fix.

## #703 — Duplicate request guard

`submitting` flag pattern added to four entry points: `LoginPage`, `RegisterPage`, `StoryCreatePage`, `RecipeCreatePage`. Each does the same thing — early-return on the handler if already in-flight, `setSubmitting(true)` before the network call, `finally { setSubmitting(false) }` to reset, plus a `disabled={submitting}` on the submit button and a "…" loading copy. Consistent shape across all four files.

One small note: `RecipeCreatePage` keeps `submitting` set to `true` on the partial-failure path (when JSON POST succeeded but PATCH failed) since the page is about to redirect — that's correct, the navigate happens 1.5s later via `setTimeout` and you don't want a quick double-click on the now-faded buttons. On the full-failure path, `setSubmitting(false)` is called in the catch block. Nice attention to the asymmetric outcomes.

## #704 — Partial-failure orphan recipe

The cleanest fix in the bunch: when `createRecipe()` succeeds but `updateRecipe(id, mediaData)` fails, you'd previously leave an orphan recipe behind with no media and no path back for the user — the page would render a generic error toast and the recipe was effectively lost. The fix:

- Clears the draft (it's now saved)
- Shows a distinct toast: "Recipe published but media upload failed — open it to retry."
- Navigates to `/recipes/<id>` after 1.5s so the user lands on their recipe and can retry media via the edit page

Exactly what you want — succeed-loudly-fail-graceful instead of silently abandoning. The 1.5s delay gives the toast time to be read.

## CI

Green (all three: Backend / Frontend / Mobile). Three-way merge against current `main` clean, only the 7 frontend pages in the commit. Approving.

PR #799, opened 2026-05-12
fix(frontend): auth race on hard refresh + linked_recipe integer crash (#694, #695)

Daglar1500:
## Summary

- **StoryEditPage, ModerationPage**: Both pages evaluated user-based guards before `AuthContext.loading` resolved. On hard refresh, `user` is `null` for a frame — story authors were redirected to the read-only detail page, and admins were sent to `/login`. Fixed by consuming `loading` from `AuthContext` and showing a loading state until auth settles.
- **StoryEditPage**: `linked_recipe` is returned as an integer pk by the backend. The page stored the raw integer in state and then read `.id` and `.title` from it (both `undefined`), causing the linked recipe label to render blank and the link to be silently cleared on every save. Fixed by resolving the pk to a full `{id, title}` object on load using the already-fetched recipes list. The save call now safely extracts the pk regardless of type.

## Test plan

 [ ] Hard-refresh `/stories/<id>/edit` as the story author — stays on the edit page, no redirect to detail
 [ ] Hard-refresh `/admin/moderation` as an admin — stays on the moderation queue, no redirect to login
 [ ] Open edit page for a story with a linked recipe — recipe title appears in the linked badge
 [ ] Save without changing the linked recipe — linked recipe survives in the API response
 [ ] Pick a new recipe from the picker, save — backend reflects the new link
 [ ] Remove the linked recipe, save — link is cleared correctly

### DIFF
--- a/app/frontend/src/pages/ModerationPage.jsx
+++ b/app/frontend/src/pages/ModerationPage.jsx
@@ -19,7 +19,7 @@ function formatDate(iso) {
 }
 
 export default function ModerationPage() {
-  const { user } = useContext(AuthContext);
+  const { user, loading: authLoading } = useContext(AuthContext);
   const navigate = useNavigate();
   const [queue, setQueue] = useState([]);
   const [loading, setLoading] = useState(true);
@@ -28,12 +28,13 @@ export default function ModerationPage() {
   const [rejectReasons, setRejectReasons] = useState({});
 
   useEffect(() => {
+    if (authLoading) return;
     if (!user) { navigate('/login'); return; }
     if (!user.is_staff) { setLoading(false); return; }
     fetchModerationQueue()
       .then(setQueue)
       .finally(() => setLoading(false));
-  }, [user, navigate]);
+  }, [user, authLoading, navigate]);
 
   const handle = async (typeKey, id, action, reason = '') => {
     setProcessing(id);
--- a/app/frontend/src/pages/StoryEditPage.jsx
+++ b/app/frontend/src/pages/StoryEditPage.jsx
@@ -11,7 +11,7 @@ import './StoryEditPage.css';
 export default function StoryEditPage() {
   const { id } = useParams();
   const navigate = useNavigate();
-  const { user } = useContext(AuthContext);
+  const { user, loading: authLoading } = useContext(AuthContext);
 
   const [story, setStory] = useState(null);
   const [title, setTitle] = useState('');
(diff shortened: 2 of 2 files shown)


ErenCanOzkaya (approved):
Two real production bugs, two clean fixes. Walking through each:

## #694 — Auth race on hard refresh

`AuthContext` exposes a `loading` flag that flips from `true` → `false` once token-bound rehydration resolves (see `AuthContext.jsx`: `useState(() => !!localStorage.getItem('token'))` initial, `setLoading(false)` in both the success and finally branches). Before this fix, `StoryEditPage` and `ModerationPage` read `user` directly and made their guard decisions immediately, while `user` was still `null` for the first render after a hard refresh. Net result:

- Story authors hitting `Cmd+R` on `/stories/:id/edit` were silently redirected to the read-only detail (the `isAuthor` check failed) and lost their edit context.
- Admins refreshing `/admin/moderation` got bounced to `/login` because the staff guard saw `user == null`.

Fix is the right shape:
```js
if (authLoading) return;
if (!user) { navigate('/login'); return; }
```

Plus `loading || authLoading` on `StoryEditPage` so the page renders the existing "Loading…" copy until both the story fetch and auth resolution finish. Dep arrays correctly updated to include `authLoading`.

## #695 — linked_recipe integer crash

Backend returns `linked_recipe` as a bare integer pk on the story detail payload. The previous code stuffed that integer into `linkedRecipe` state and later did `linkedRecipe.id` for both display and save — `.id` on an integer is `undefined`, so the linked-recipe pill rendered empty and the save call sent `undefined` to the server.

Fix resolves the integer to a full `{ id, title }` object on load using the already-fetched `recipes` list:

```js
const linkedId = storyData.linked_recipe;
if (linkedId) {
  const found = recipes.find((r) => r.id === linkedId);
  setLinkedRecipe(found ?? { id: linkedId, title: `Recipe #${linkedId}` });
} else {
  setLinkedRecipe(null);
}
```

The `?? { id, title: 'Recipe #N' }` fallback is the right call for the case where the linked recipe isn't in the user's accessible list (e.g. private to a different author) — keeps the UI honest about the link without crashing. On save the new typeof-guard handles both shapes:

```js
const recipeId = typeof linkedRecipe === 'object' ? linkedRecipe.id : linkedRecipe;
formData.append('linked_recipe', recipeId);
```

This is also forward-compat — if a future patch stops resolving and passes the raw integer, save still works.

## CI

Green (all three: Backend / Frontend / Mobile). Three-way merge against current `main` clean, only `ModerationPage.jsx` (+1/-1) and `StoryEditPage.jsx` (+15/-5) actually change. Approving.

PR #801, opened 2026-05-12
fix(frontend): MapPage crash/race/error handling + RecipeDetailPage 4 bugs (#697, #698)

Daglar1500:
## Summary

**MapPage (#697)**
- `content_count.recipes` / `.stories` crashed when `content_count` was null — replaced with `?.` optional chaining and `?? 0` fallback.
- Rapid region clicks caused a stale fetch to overwrite newer content — a `currentRegionId` ref now discards responses that no longer match the selected region.
- Region list fetch had no `.catch()` — failure rendered a blank map with no feedback. Added error state with a visible message.
- Per-region content fetch had no `.catch()` — failure was silent. Content is now cleared and loading ends cleanly on error.

**RecipeDetailPage (#698)**
- Navigating from recipe A to recipe B carried over A's check-offs, shopping list, and substitutions — added a `useEffect` on `id` that resets all per-recipe state.
- `recipe.ingredients` being null/undefined crashed the page — guarded with `?? []` in all three usages.
- Same ingredient appearing twice caused React duplicate-key warning — key is now `ingredient-index`.
- `toUsername` in the message compose link was not URL-encoded — usernames with spaces or unicode broke the route.

## Test plan

 [ ] Open Map page, kill the backend — clear error message appears instead of a blank screen
 [ ] Click region A, immediately click region B — B's content renders (not A's)
 [ ] Click any region whose `content_count` is null — panel renders without crash
 [ ] Open recipe A, check off ingredients, navigate to recipe B — B starts with a clean shopping list
 [ ] Open a recipe with `ingredients: null` — page renders without crash
 [ ] Open a recipe whose author has a unicode/space username, click Message — compose URL is valid

### DIFF
--- a/app/frontend/src/pages/MapPage.jsx
+++ b/app/frontend/src/pages/MapPage.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { Link } from 'react-router-dom';
 import { MapContainer, TileLayer, CircleMarker, Tooltip } from 'react-leaflet';
 import 'leaflet/dist/leaflet.css';
@@ -10,7 +10,9 @@ export default function MapPage() {
   const [selected, setSelected] = useState(null);
   const [content, setContent] = useState([]);
   const [loading, setLoading] = useState(true);
+  const [loadError, setLoadError] = useState('');
   const [contentLoading, setContentLoading] = useState(false);
+  const currentRegionId = useRef(null);
 
   useEffect(() => {
     fetchMapRegions()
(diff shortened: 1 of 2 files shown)


ErenCanOzkaya (approved):
Eight real fixes across two pages, all the right shape and the right size. Walking through:

## #697 — MapPage (4 fixes)

**Stale fetch race** — the `currentRegionId` ref pattern is exactly right: capture the region id at fetch-start, gate every state update on `currentRegionId.current === region.id` so a slow response for an abandoned region is silently dropped. The same gate covers `.then`, `.catch`, and `.finally` consistently — no path can flicker `setContent` or `setContentLoading` for a region the user has already moved past.

**Regions fetch error state** — previously a failed `fetchMapRegions()` left the page hanging on `loading: true` after `.finally` cleared it (no error message rendered). Now `loadError` surfaces a "Could not load regions." string in the map slot. `.map-error` CSS class added (would have flagged if it didn't exist — verified the corresponding rule lives in `MapPage.css` or `loadError` would render unstyled).

**Content fetch error state** — symmetrical fix. On `.catch()` the content is cleared (so a partial display from a previous region doesn't linger) and loading flips off cleanly.

**`content_count` null guard** — `selected.content_count?.recipes ?? 0` covers the case where the backend ever omits the counts subobject. Cheap, defensive.

## #698 — RecipeDetailPage (4 fixes)

**State leak A → B** — the new `useEffect` keyed on `[id]` resets `checked`, `showShoppingList`, `openSubPanel`, `substitutes`, `appliedSubs` whenever the route param flips. This is the only effect that owns those resets, so the existing `fetchRecipe` effect (also `[id]`) doesn't need any change — React guarantees declaration order on co-mounted effects, so reset runs before the new load resolves and overwrites state. Nice clean split.

**`recipe.ingredients` null guard** — three call sites (`hasConverted`, `shoppingItems`, the JSX `.map`) all wrapped in `(recipe.ingredients ?? [])`. Matches the defensive shape used elsewhere on the page.

**Duplicate key fix** — `${ri.ingredient}-${index}` composite key. The index alone would be sufficient since it's already unique within the list; the composite is just clearer at a glance about what makes the row unique. Either way, React's duplicate-key warning is gone.

**URL-encode `toUsername`** — `recipe.title` was already wrapped in `encodeURIComponent`, so adding it to `author_username` brings the link to a consistent state. Usernames with spaces, parentheses, or unicode (e.g. Turkish characters) now produce a valid URL. The `to=${recipe.author}` is a numeric id so no encoding needed there.

## CI

Green across all three checks (Backend / Frontend / Mobile). Three-way merge against current `main` clean, only the two listed frontend pages actually change. Approving.

PR #804, opened 2026-05-12
feat(mobile): EditProfileScreen — display name, bio, avatar (if backend supports), prefs link (closes #661)

CaganIslam:
## Summary
- Adds a real `EditProfileScreen` (mobile) wired into the Profile tab's "Edit profile" row, replacing the previous "Coming soon" toast.
- New `profileService.ts` exposes `fetchOwnProfile` (GET) and `updateOwnProfile` (PATCH) against `/api/users/me/`, matching the existing `UserProfileSerializer` / `UserPreferencesUpdateSerializer` shape.
- Form pre-fills from a fresh GET on mount; saves only changed fields via a partial PATCH, mirrors the update into `AuthContext` via `updateUser`, shows a success toast, and navigates back.
- Username and email are read-only (with "Contact support to change" hint); cultural / dietary preferences are routed through the existing Onboarding screen via a "Edit cultural preferences" link.
- Discard-confirm dialog on back button when the form is dirty; bio capped at 500 chars with a live counter; region capped at the backend's 100-char limit.

## Backend endpoint
- `PATCH /api/users/me/` (DRF; `UserPreferencesUpdateSerializer`).
- Fields wired up in this PR: **`bio`** (multiline, 500-char client cap), **`region`** (100-char cap). Username, email, role, and `created_at` are read-only per serializer. Cultural / regional / religious / event preferences are intentionally delegated to the existing Onboarding flow to avoid duplicating the chip-picker UI.
- **Avatar upload was NOT shipped:** the `User` model (`app/backend/apps/users/models.py`) has no avatar / image field and no `/api/users/me/avatar/` endpoint exists. The screen omits avatar entirely; can be added in a follow-up once the backend exposes it.
- No `display_name` field on the backend either — the spec mentioned it, but the model only has `username`, so display-name editing is also deferred.

## Test plan
- [ ] Log in as `ayse@example.com` / `StrongPassword123!`, open Profile tab → tap "Edit profile" → screen pre-fills with the user's bio and region.
- [ ] Change bio to a new string, tap Save → success toast appears, screen pops back, ProfileTab still shows the same `@username`.
- [ ] Type more than 500 chars into bio → counter turns red, Save is disabled.
- [ ] Edit a field, tap back arrow → discard confirm dialog appears; tap "Keep editing" returns to form, "Discard" pops back without saving.
- [ ] Tap "Edit cultural preferences" → routes to the existing Onboarding screen.
- [ ] Username and email rows are visibly read-only with the support hint.

## Quality gates
- `npx tsc --noEmit` — clean.
- `npx jest` — **155/155 passing** (26 suites), including:
  - `__tests__/services/profileService.test.ts` (3 cases: GET path, PATCH with changed fields, empty patch).
  - `__tests__/screens/EditProfileScreen.test.tsx` (2 cases: renders fetched profile, save sends only changed fields and navigates back).
- Metro bundle (`/index.bundle?platform=ios&dev=false&minify=true`) returned HTTP 200.

Closes #661

### DIFF
--- a/app/mobile/__tests__/screens/EditProfileScreen.test.tsx
+++ b/app/mobile/__tests__/screens/EditProfileScreen.test.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import { render, fireEvent, waitFor, act } from '@testing-library/react-native';
+import { SafeAreaProvider } from 'react-native-safe-area-context';
+
+jest.mock('../../src/services/profileService', () => ({
+  fetchOwnProfile: jest.fn(),
+  updateOwnProfile: jest.fn(),
+}));
+
+const mockUpdateUser = jest.fn().mockResolvedValue(undefined);
+jest.mock('../../src/context/AuthContext', () => ({
+  useAuth: () => ({
+    user: { id: '1', username: 'ayse', email: 'ayse@example.com' },
+    updateUser: mockUpdateUser,
+  }),
+}));
+
+const mockShowToast = jest.fn();
+jest.mock('../../src/context/ToastContext', () => ({
+  useToast: () => ({ showToast: mockShowToast }),
+}));
+
+import EditProfileScreen from '../../src/screens/EditProfileScreen';
+import {
+  fetchOwnProfile,
+  updateOwnProfile,
+} from '../../src/services/profileService';
+
+const mockedFetch = fetchOwnProfile as jest.MockedFunction<typeof fetchOwnProfile>;
+const mockedUpdate = updateOwnProfile as jest.MockedFunction<typeof updateOwnProfile>;
+
+function makeProps() {
+  const navigation = {
+    navigate: jest.fn(),
+    goBack: jest.fn(),
+    setOptions: jest.fn(),
+    addListener: jest.fn(() => () => undefined),
+  } as any;
+  const route = { key: 'k', name: 'EditProfile', params: undefined } as any;
+  return { navigation, route };
+}
+
+function renderScreen() {
+  const props = makeProps();
+  const utils = render(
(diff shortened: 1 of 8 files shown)


ErenCanOzkaya (approved):
LGTM — clean addition.

## What I verified
- `profileService` hits the right backend route (`/api/users/me/`, matching `UserProfileSerializer` / `UserPreferencesUpdateSerializer`).
- Partial PATCH semantics: only changed fields are sent (the test asserts exactly that — `{ bio: 'New bio' }` and nothing else).
- `updateUser` mirror back to `AuthContext` preserves existing values for any field the response omits (`updated.X ?? user.X`) — won't silently null cultural preferences.
- Navigation: `EditProfile` is registered in `PublicStackNavigator`, and `RootTabsNavigator` mounts `PublicStackNavigator` under the `Feed` tab — so `navigation.navigate('Feed', { screen: 'EditProfile' })` from `ProfileTab` is correct.
- Discard-on-back guard fires only when the form is dirty (`isDirty` memo). Bio counter goes red and `canSave` flips false when over the 500-char cap; region is hard-capped to 100 chars at input time (matches backend).
- Username / email correctly read-only with the support hint — backend serializer marks them read-only.
- Avatar / display_name deferral is documented and correct (no field on the `User` model yet).
- All three CI checks green; 26 jest suites / 155 tests pass per PR body.

## Minor (non-blocking)
- The `if (user)` guard around `updateUser` in `handleSave` is fine, but if the user is null at save time, the save still succeeded server-side while AuthContext won't know — vanishingly rare (you can't reach this screen unauthenticated), so not worth a guard.

Approving.

PR #808, opened 2026-05-12
feat(backend): seed a richer mock 1:1 conversation for messaging UI (#795)

laternis:
Summary
seed_test_db (seed_messaging): The seeded messaging flow only created a single placeholder message between `cook` and `student`, which wasn't enough to test inbox ordering, last-message preview, thread rendering or reply flow. Replaced it with a realistic 7-turn back-and-forth grounded in the seeded "Stuffed Eggplant (Karnıyarık)" recipe (vegetarian substitution → roasting time → region tag), with deterministic timestamps (fixed base + constant per-turn offset).

seed_test_db (idempotency): The thread is now matched by *both* participants, messages are created via `get_or_create`, and any stale messages left over from earlier seed versions (e.g. the old "Hello student!" placeholder) are pruned. The thread's `last_message_at` / `last_message_preview` are recomputed from the newest message. Re-running the seed produces identical data — no duplicate threads or messages.

Related: #795

Test plan
 [ ] Run `python manage.py seed_test_db`; inbox endpoint shows one cook↔student thread with the preview matching the newest message
 [ ] Open the thread — 7 alternating messages in chronological order, not a single placeholder
 [ ] Re-run the seed — still exactly one thread and 7 messages (no duplicates)
 [ ] If a thread with an old single placeholder message already exists, re-running replaces it with the full conversation

### DIFF
--- a/app/backend/apps/common/management/commands/seed_test_db.py
+++ b/app/backend/apps/common/management/commands/seed_test_db.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
+from datetime import datetime, timedelta
+
 from django.core.management.base import BaseCommand
 from django.contrib.auth import get_user_model
+from django.utils import timezone
 from apps.recipes.models import Region, Ingredient, Unit, Recipe, RecipeIngredient
 from apps.stories.models import Story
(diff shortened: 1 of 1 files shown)


roboticrustacean (approved):
awesome work

PR #812, opened 2026-05-12
fix(frontend): error handling, mock guards, region endpoint (#699 #700 #701)

Daglar1500:
## Summary

- **RecipeCommentsSection**: `handleDelete` surfaces a visible inline error (`role="alert"`) when `deleteComment` fails instead of swallowing the rejection.
- **IngredientRow**: `handleAddIngredient` / `handleAddUnit` show inline `role="alert"` errors when the parent's create call rejects (was silent catch).
- **ModerationPage**: added `fetchError` (queue load) and `actionError` (approve/reject) states with visible error messages.
- **EventDetailPage**: reset tab to `'all'` when `eventId` param changes to avoid stale tab after navigation.
- **searchService**: `fetchRegions` calls `/api/map/regions/?geo_only=true` (only regions with geo coordinates, which is what the map-backed region picker needs). `search` now sends `meal_type` as a server-side filter param (#700).
- **SearchPage**: drops the client-side `meal_type` substring fallback now that the backend (`#856`) supports the field. Keeps a small client-side `story_type` filter for story results until the unified search endpoint plumbs the param through.
- **storyService**: added USE_MOCK guard to `updateStory`.
- **mapService**: added USE_MOCK guard to `fetchMapRegionContent`.
- **notificationService**: added USE_MOCK guard to `markAllAsRead`.
- **api.js**: silent JWT refresh dispatches `auth:token-refreshed` custom event after writing to localStorage.
- **AuthContext**: listens for `auth:token-refreshed` to sync React token state with localStorage after silent refresh.

closes #700
closes #701

refs #699 — this PR addresses RecipeCommentsSection / IngredientRow / EventDetailPage / ModerationPage from the issue scope. StoryCreatePage, StoryDetailPage and RecipeCreatePage already have proper error surfaces (Toast / `deleteError` / `publishError` / `storySaveError` / form-level error state). A follow-up issue will track any additional polish on those pages so #699 stays open until each named page is independently verified.

## Test plan

- [ ] Open a recipe with comments, delete a comment with a failing network — confirm inline "Could not delete comment" alert appears.
- [ ] On recipe create/edit, type a new ingredient and tap Add with a failing network — confirm inline "Could not add" error.
- [ ] On Moderation page with a failing fetch — confirm visible error banner.
- [ ] Navigate between two event detail pages — confirm tab resets to "All" on each.
- [ ] On search page, region filter populates — verify it calls `/api/map/regions/?geo_only=true`.
- [ ] Search with `?meal_type=breakfast` — verify the request to `/api/search/` includes `meal_type=breakfast`.
- [ ] With `REACT_APP_USE_MOCK=true`: edit and save a story / open Map page / mark all notifications as read — confirm no network errors.
- [ ] Let access token expire, make a request — confirm silent refresh updates both localStorage and React AuthContext token state.


ErenCanOzkaya (changes requested):
Request changes — approach is right, but the diff under-delivers vs. the three issues it claims to close. Verdict per issue:

### #701 — mock guards + refresh → AuthContext sync ✅
- Early `if (USE_MOCK) return …` guards on `storyService.updateStory`, `mapService.fetchMapRegionContent`, `notificationService.markAllAsRead` — good.
- `api.js` dispatches `auth:token-refreshed` after writing localStorage and `AuthContext` listens + updates state, with proper unsubscribe on unmount. Wiring is clean.

### #700 — searchService region endpoint + meal_type ⚠️ half-done
- Region endpoint fix: changed from `/api/regions/` to `/api/map/regions/` with `geo_only` param — good, BUT the PR description says `geo_only=false` while the code sends `geo_only: true`. Which is intended? They have opposite semantics.
- **`meal_type` is not addressed.** `SearchPage.jsx` lines ~132–134 still filter client-side via `r.title.toLowerCase().includes(mealType...)`, and `searchService.search` doesn't add `meal_type` to params. That's half of #700 missing — please move the filter server-side or remove the issue mention from the PR description.

### #699 — error handling + UX polish ⚠️ partial + silent catches
- Touched: `RecipeCommentsSection.handleDelete`, `IngredientRow.handleAddIngredient/Unit`, `EventDetailPage` tab reset.
- **Missing entirely:** `ModerationPage`, `StoryCreatePage`, `StoryDetailPage`, `RecipeCreatePage` — all named in #699's scope. Auto-closing the issue on merge would leave real polish work undone.
- **Silent catches that don't surface anything to the user:**
  - `IngredientRow.jsx`: empty `catch {}` with a comment saying the parent surfaces the error — but the parent only sees a resolved `undefined`; if `onNewIngredient` rejects, the rejection is dropped here before it can bubble.
  - `RecipeCommentsSection.jsx` delete failure: catch only closes the menu; no toast / inline error. User has no idea the delete failed.
  These match "no crash" but #699 asks for actual user feedback.

### Tests
- `searchService.test.js` updated for the new region endpoint — good.
- Missing: no test for the `auth:token-refreshed` event (dispatch or listener), no tests for the new mock guards (`updateStory`, `fetchMapRegionContent`, `markAllAsRead` under `USE_MOCK=true`), no failure-path test for `RecipeCommentsSection.handleDelete` or `IngredientRow` rejection.

### Asks before merge
1. Resolve the `geo_only` description-vs-code discrepancy.
2. Address `meal_type` (server-side filter or scope it out of the PR description).
3. Either cover the four remaining #699 pages or remove #699 from the closes list.
4. Surface delete/add failures as user-visible errors in `RecipeCommentsSection` + `IngredientRow` instead of swallowing.

Happy to re-review once these are in.

ErenCanOzkaya (changes requested):
Big progress on this round — \`meal_type\` is server-side now, the failure-surfacing in \`RecipeCommentsSection\` / \`IngredientRow\` reads well with the \`role=\"alert\"\` lines, and the test coverage gaps from last review (token-refresh, mock guards, failure paths) are filled in cleanly. Three things still standing between this and approve.

### 1. \`geo_only\` body / code / test still disagree
PR description says \`/api/map/regions/?geo_only=false\`, but \`searchService.js\` sends \`{ params: { geo_only: true } }\` and the test asserts \`geo_only: true\`. The sign flipped vs. last review but the inconsistency is still there. Pick one and make all three agree — \`true\` keeps the geo-only behaviour I'd expect for the region filter, \`false\` returns all regions including non-geo ones; whichever you actually want, just align them.

### 2. \`closes #699\` is still premature
Issue #699's scope lists Moderation, RecipeComments, IngredientRow, StoryCreate, StoryDetail, EventDetail, RecipeCreate. This round added ModerationPage error handling (good), but **StoryCreatePage**, **StoryDetailPage**, and **RecipeCreatePage** still aren't touched by the diff. Two options:
- Add the missing three (probably similar pattern — \`role=\"alert\"\` for the submit/load failure path), keep \`closes #699\`.
- Drop \`closes #699\` from the description, open a follow-up issue for the three remaining pages and \`closes #<that>\` instead. Either is fine; please don't auto-close #699 on merge while three of its named pages are still unhandled.

### 3. Branch needs a rebase
Mergeable state is \`DIRTY\`. Four files conflict with current \`main\`:
- \`app/frontend/src/services/mapService.js\`
- \`app/frontend/src/services/storyService.js\`
- \`app/frontend/src/pages/SearchPage.jsx\`
- \`app/frontend/src/__tests__/mapService.test.js\`

Likely the recently merged contract micro-cleanups + paginated-fetch PRs landed on the same surface. Rebase/merge \`main\` and resolve.

### Minor (not blocking)
- \`ModerationPage\` got new error/UX state in this PR but no \`ModerationPage.test.jsx\` change — a small test covering the \`fetchError\` and \`actionError\` branches would round it out, but not blocking.

Happy to re-review and approve once 1–3 are addressed.

ErenCanOzkaya (approved):
Round 3 looks good — all three blocking items resolved:

- **\`geo_only\` consistency** ✅ Body, code, and test all assert \`geo_only: true\` now.
- **\`closes #699\` scope** ✅ Body switched to \`refs #699\` with the explicit note that StoryCreate / StoryDetail / RecipeCreate already have error surfaces, and #699 stays open for any follow-up polish. Right call.
- **Merge conflicts** ✅ No conflict markers in the diff; \`BLOCKED\` is now just the review gate.

Test coverage round-up reads well — \`searchService\` / \`AuthContext\` / \`notificationService\` / \`storyService\` / \`mapService\` / \`RecipeCommentsSection\` / \`IngredientRow\` all got matching test updates this round, and \`SearchPage.test.jsx\` correctly swapped the old client-side \`meal_type\` substring test for a server-side parameter assertion.

**Non-blocking nit:** \`ModerationPage\` gained \`fetchError\` / \`actionError\` rendering but no \`ModerationPage.test.jsx\` covers those branches yet. Worth a small follow-up.

LGTM — ready to merge.

PR #815, opened 2026-05-12
feat(mobile): render and edit recipe steps — detail section + create/edit repeater (closes #806)

CaganIslam:
## Summary
- Mobile now surfaces backend `Recipe.steps` (JSONField of strings) — Detail screen renders a numbered "Steps" section between description and ingredients, hiding itself entirely when the array is empty / null / undefined.
- Create and Edit forms get a shared `RecipeStepsRepeater` (multiline input per step, ▲/▼ reorder, ✕ remove, "Add step" pill) placed between description and ingredients. Auto-renumbers as rows shift.
- Steps state is trimmed on submit (`trimStepsForPayload`) so empty rows are dropped; the empty list is a valid payload.
- Edit pre-fills rows from `recipe.steps` and clears to empty when the field is missing.
- Adds `steps?: string[]` to `RecipeDetail` and threads it through `buildRecipePatchJsonBody` so the field rides the existing JSON PATCH (not multipart) — see notes below.

## Payload path
- **Create**: `apiPostJson('/api/recipes/', { …, steps })` — single JSON POST, same call that already carries `ingredients_write`.
- **Edit**: `patchRecipeJson(id, body)` where `body` comes from `buildRecipePatchJsonBody({ …, steps })`. `Recipe.steps` is a JSONField, so it rides the JSON PATCH alongside `ingredients_write` — multipart is reserved for the image / video uploads, which still run as independent follow-up PATCHes. When the caller omits `steps`, the builder leaves the key off the body so the server value is preserved; when the user clears every row, `steps: []` is sent explicitly.

## Test plan
- [ ] Open a recipe with non-empty steps on mobile — verify the numbered "Steps" section appears between description and ingredients, with mustard circle badges and wrapping body text.
- [ ] Open a recipe with no `steps` (or empty array) — verify the Steps section is absent (no "No steps yet" placeholder).
- [ ] Create a new recipe with 3 steps, reorder one up, delete one, leave one row blank — submit and verify the created recipe shows the surviving trimmed steps in order.
- [ ] Edit an existing recipe — verify existing steps pre-fill, reorder works, save persists.
- [ ] Edit and clear every step row, save, reload — verify the section disappears on the detail screen.

## Verification
- `npx tsc --noEmit` clean.
- `npx jest` — **166 tests passing across 27 suites** (baseline ~155; +11 new). New suites: `RecipeStepsSection`, `RecipeStepsRepeater`, `recipeServiceSteps`.
- Metro `index.bundle?platform=ios&dev=false&minify=true` returns HTTP 200.

Closes #806

### DIFF
--- a/app/mobile/__tests__/components/RecipeStepsRepeater.test.tsx
+++ b/app/mobile/__tests__/components/RecipeStepsRepeater.test.tsx
@@ -1,0 +1,111 @@
+import React, { useState } from 'react';
+import { fireEvent, render } from '@testing-library/react-native';
+import {
+  RecipeStepsRepeater,
+  trimStepsForPayload,
+} from '../../src/components/recipe/RecipeStepsRepeater';
+
+/**
+ * Tiny harness so we can drive the controlled repeater through React state
+ * and assert what the consumer ultimately holds — closer to how the screens
+ * actually use it than re-rendering with a fresh `steps` prop each tap.
+ */
+function Harness({
+  initial,
+  onCommit,
+}: {
+  initial: string[];
+  onCommit: (steps: string[]) => void;
+}) {
+  const [steps, setSteps] = useState<string[]>(initial);
+  return (
+    <RecipeStepsRepeater
+      steps={steps}
+      onChange={(next) => {
+        setSteps(next);
+        onCommit(next);
+      }}
+    />
+  );
+}
+
+describe('trimStepsForPayload', () => {
+  it('trims whitespace and drops empty rows', () => {
+    expect(trimStepsForPayload(['  one  ', '', '\n\t', 'two'])).toEqual(['one', 'two']);
+  });
+
+  it('returns an empty array when nothing survives the trim', () => {
+    expect(trimStepsForPayload(['', '   ', '\n'])).toEqual([]);
+  });
+});
+
+describe('RecipeStepsRepeater', () => {
+  it('renders the empty hint and no rows when steps is empty', () => {
+    const onCommit = jest.fn();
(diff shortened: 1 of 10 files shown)


ErenCanOzkaya (approved):
LGTM — well-factored.

## What I verified
- Three-way payload semantics on `buildRecipePatchJsonBody` are correct and tested:
  - `steps` omitted from input → key not in body (server value preserved).
  - `steps: []` passed → `steps: []` sent (explicit clear).
  - `steps: [...]` passed → forwarded as-is.
- `trimStepsForPayload` is pure, exported, unit-tested (whitespace-only rows dropped, all-empty produces `[]`).
- `RecipeStepsRepeater` is fully controlled — state lives in the parent, all mutations through `onChange`, which lets Create and Edit share the component without coupling. Tests cover add / remove / reorder up + down / first-row up-disabled / last-row down-disabled / text edit.
- `RecipeStepsSection` correctly returns `null` for empty / undefined arrays (no "No steps yet" placeholder, matches the description block's behaviour) and preserves embedded newlines.
- Edit screen pre-fills via `Array.isArray(recipe.steps) ? recipe.steps.slice() : []` (defensive against missing field) and submits `trimStepsForPayload(steps)` — so leaving every row blank ships `[]` and clears server-side, as the PR body promises.
- `RecipeDetail.steps?: string[]` typed as optional — won't break existing call sites.
- All three CI checks green; 27 jest suites / 166 tests pass per PR body.

## Minor (non-blocking)
- Reorder controls are ▲/▼ buttons rather than drag — fine for v1; if step lists grow, drag-to-reorder would scale better. Not in scope here.

Approving.

PR #816, opened 2026-05-12
fix(frontend): timer leaks, keyboard shortcut re-register, notification cascade renders (#702)

Daglar1500:
## Summary

- **useKeyboardShortcuts**: listener was re-registered on every render because callers pass a new object literal each time. Now uses a ref to always hold the latest shortcuts map; the listener registers exactly once.
- **NotificationContext**: `markRead` had `notifications` in its `useCallback` deps, causing the callback to be re-created on every notification state update and cascading re-renders. Fixed by reading the current notifications via a `notificationsRef` instead.
- **RecipeCreatePage / RecipeEditPage**: `setTimeout` IDs for toast auto-dismiss and post-save navigation were not stored, so they couldn't be cleared on unmount. Both now use `toastTimerRef` and `navTimerRef`, cleared in a cleanup `useEffect`.

references #702

## Test plan

- [ ] Open React DevTools profiler, mark a notification as read — confirm no cascade re-renders from `NotificationContext`
- [ ] Type while focused on an input on any page — confirm keyboard shortcuts don't fire
- [ ] Create a recipe, immediately navigate away before the 1.5 s nav timer fires — no "Can't perform a React state update on an unmounted component" warning in console
- [ ] Edit a recipe, same unmount-during-timer test
- [ ] Verify keyboard shortcuts still work normally when no input is focused

### DIFF
--- a/app/frontend/src/context/NotificationContext.jsx
+++ b/app/frontend/src/context/NotificationContext.jsx
@@ -17,6 +17,8 @@ export function NotificationProvider({ children }) {
   const [notifications, setNotifications] = useState([]);
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState('');
+  const notificationsRef = useRef([]);
+  useEffect(() => { notificationsRef.current = notifications; }, [notifications]);
 
   const refreshNotifications = useCallback(async () => {
     if (!token) {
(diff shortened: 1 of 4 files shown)


ErenCanOzkaya (approved):
LGTM — fixes look solid:

- **Timer leaks (RecipeCreate/EditPage):** `setTimeout`s now held in `toastTimerRef` / `navTimerRef`, cleared on unmount, and `showToast` pre-clears the prior timer. Clean.
- **Keyboard shortcut re-register:** the ref-mirror pattern (`shortcutsRef` updated every render, listener attached once with `[]` deps) is the right shape.
- **Notification cascade renders:** dropping `notifications` from `markRead`'s deps via `notificationsRef` gives a stable callback identity — no more cascade.

One thing worth confirming with the assignee: issue #702 also lists "duplicate React keys" as a subarea, which this PR doesn't touch. Either intentionally deferred to another PR, or the issue should be split — either way, please don't auto-close #702 on merge if the key work is still outstanding.

A unit test on `useKeyboardShortcuts` verifying single `addEventListener` across renders would be cheap and high-value, but not blocking.

PR #818, opened 2026-05-12
fix(backend): repair drifted seed_test_db command (#814)

laternis:
Summary
seed_test_db (seed_recipes): The seeded recipes referenced a measurement unit named `units`, which was renamed to `pieces` when units moved into the `0005_seed_ingredients_units` data migration. `Unit.objects.get(name='units')` raised `Unit.DoesNotExist`, and since `handle()` wraps everything in `transaction.atomic()` the entire seed (including the messaging conversation) rolled back. Pointed the Tomato and Onion ingredients at `pieces`.

seed_test_db (seed_stories): The command passed a `linked_recipe` kwarg to `Story`, but that FK is now a `linked_recipes` ManyToManyField (with a separate `region` FK). Dropped `linked_recipe` from the create defaults and attach the recipe via `story.linked_recipes.add()` after the story row exists.

Closes #814

Test plan
 [ ] On a fresh DB, `python manage.py migrate && python manage.py seed_test_db` runs to completion ("Successfully seeded database!")
 [ ] Re-run `python manage.py seed_test_db` — no errors, every record reported as "already exists" (idempotent)
 [ ] Inspect the seeded data: both recipes have their RecipeIngredients with correct units (Tomato 2 pieces, Garlic 3 cloves, ...); the seeded story is linked to a recipe via `linked_recipes`; the cook↔student thread still has its 7-message conversation

### DIFF
--- a/app/backend/apps/common/management/commands/seed_test_db.py
+++ b/app/backend/apps/common/management/commands/seed_test_db.py
@@ -140,7 +140,7 @@ class Command(BaseCommand):
                 'region': med_region,
                 'is_published': True,
                 'ingredients': [
-                    {'name': 'Tomato', 'amount': 2, 'unit': 'units'},
+                    {'name': 'Tomato', 'amount': 2, 'unit': 'pieces'},
                     {'name': 'Garlic', 'amount': 3, 'unit': 'cloves'},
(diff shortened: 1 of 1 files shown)


roboticrustacean (approved):
thanks for the fix

PR #820, opened 2026-05-12
feat(frontend): profile page shows my recipes, stories, and saved recipes (#709)

Daglar1500:
## Summary

The `/profile` page previously showed only username, email, a contactability toggle, and a logout button. This PR adds a full content section with three tabs.

- **My Recipes** — `GET /api/recipes/?author=<userId>` rendered as a card grid with count in the tab label
- **My Stories** — `GET /api/stories/?author=<userId>` same layout
- **Saved Recipes** — `GET /api/recipes/?bookmarked=true`; relies on the bookmark API already shipped in backend
- Each tab shows a thumbnail + title card linking to the detail page; friendly empty state when there's nothing yet
- `recipeService`: new `fetchMyRecipes(userId)` and `fetchBookmarkedRecipes()`, both with USE_MOCK guards; `fetchRecipes` now accepts optional `params`
- `storyService`: new `fetchMyStories(userId)`

closes #709

## Test plan

- [ ] Log in, go to `/profile` — all three tabs appear with counts
- [ ] "My Recipes" tab shows your published recipes as cards, each links to its detail page
- [ ] "My Stories" tab shows your published stories
- [ ] "Saved Recipes" tab shows bookmarked recipes (bookmark a recipe on its detail page first)
- [ ] Empty state copy shows when a section has no items
- [ ] Tabs switch without page reload


ErenCanOzkaya (changes requested):
Request changes — issue #709 was already closed by PR #843 (merged 2026-05-12), so this PR overlaps that work. The tabbed layout here is a real UX choice vs. main's 3-column dashboard, but the diff also drops a couple of safeguards that were on main:

- **Promise.allSettled → three independent then/catch/finally chains.** Failed fetches now resolve to empty arrays and look identical to "no content" — the per-section error state on main (`recipesError` / `storiesError` / `bookmarksError`) is gone. The pre-existing "could not load stories" test was renamed to "you haven't published any stories yet" — that's removing error-state coverage, not migrating it.
- **No `cancelled` flag on the fetch effect.** Main has `let cancelled = false` cleanup; the new effect can `setState` after unmount during rapid navigation. Please restore.
- **`role="tablist"` pattern is incomplete.** Tab buttons have `role="tab"` + `aria-selected` but there are no matching `role="tabpanel"`s, no `aria-controls` / `id` linkage, and no left/right arrow keyboard handling. Either finish the pattern (WAI-ARIA tablist) or drop the roles and style as plain buttons.
- **`fetchRecipes(params)` signature is widened in `recipeService.js` but unused by this PR's diff.** Either wire it up or revert that change so we don't ship dead surface area.

Two options:
1. Keep this PR open if the tabbed UX is intentionally replacing the dashboard — restore error states + cancelled flag, finish the tablist a11y, drop the unused `fetchRecipes(params)` widening.
2. If it's superseded by #843, please close this PR + branch.

Either way is fine; I'll re-review once it's clear which direction this is taking.

PR #828, opened 2026-05-12
fix(mobile): smoother region camera animation, regional zoom level, drop sheet flicker on Map Discovery (closes #825)

CaganIslam:
## Summary
Map Discovery: tapping a pin used to flash the region bottom sheet for a few frames before the next screen pushed, the camera snapped in too fast, and it zoomed in so tight you lost all regional context. This made the transition feel jerky and disorienting.

- Removed the `setFocused(pin)` call from the pin `onPress` handler so the bottom sheet no longer mounts/unmounts during navigation (kills the flicker).
- Swapped the camera target from `zoom: 5` / `duration: 450ms` to `zoom: 9` / `duration: 700ms` — lands at a roughly regional 10 km frame instead of dropping the user onto the pin.
- Delayed `navigation.navigate('RegionMapDetail', ...)` by 700 ms via `setTimeout` so the camera animation visibly settles before the next screen pushes. Timeout id is stored in a ref and cleared on unmount.

**Camera approach:** went with `animateCamera({ center, zoom: 9 }, { duration: 700 })` — same API that was already in use, just retuned. Did not need to fall back to `animateToRegion`.

## Test plan
- [ ] Open Map Discovery, tap any region pin → camera animates smoothly for ~0.7s to a regional view (~10 km), then RegionMapDetail pushes.
- [ ] Confirm the bottom RegionDetailSheet does NOT briefly appear during the tap-then-navigate sequence.
- [ ] Tap a pin and immediately background/back out before 700 ms — no crash, no orphan navigation push.
- [ ] Tapping empty map still dismisses any open sheet (regression check).
- [ ] 🧭 Ingredient routes pill, zoom controls, and Search regions pill still behave as before.

Closes #825

### DIFF
--- a/app/mobile/src/screens/MapDiscoveryScreen.tsx
+++ b/app/mobile/src/screens/MapDiscoveryScreen.tsx
@@ -15,10 +15,15 @@ import { shadows, tokens, useTheme } from '../theme';
 
 type Props = NativeStackScreenProps<RootStackParamList, 'MapDiscovery'>;
 
-/** Slight zoom-in when a pin is selected. react-native-maps' `animateCamera`
- * uses Google Maps zoom levels (higher = closer); 5 lands roughly at a country
- * frame, which feels good after the global initial view. */
-const FOCUS_ZOOM = 5;
+/** Regional zoom-in when a pin is selected. react-native-maps' `animateCamera`
+ * uses Google Maps zoom levels (higher = closer); zoom 9 lands at a roughly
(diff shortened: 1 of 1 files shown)


mustafaocakxyz (approved):
Good fix for #825: deferring navigation until the camera settles and skipping the transient focused/sheet state removes the flicker. Cleanup on unmount avoids navigating after unmount. 

Important to remember: rebase after #830 (or land first) since both touch MapDiscoveryScreen. 

LGTM.

PR #830, opened 2026-05-12
fix(mobile): Search regions pill visibility on MapDiscoveryScreen

CaganIslam:
## Summary
The pill JSX existed on `main` from PR #779 but was wrapped in a `pointerEvents="box-none"` parent at the same z-order as the map overlay — easy to miss visually, possibly hidden behind the MapView on some Android renders.

Simplified: removed the wrapper, applied `position: absolute, top: 12, alignSelf: 'center', zIndex: 10` directly on the Pressable. Now sits unambiguously above the map.

Closes #826


mustafaocakxyz (approved):
Layout fix makes the Search regions CTA reliably visible above the map (z-index + absolute positioning). LGTM — merge order / rebase with #828 may be needed since both edit the same screen.

PR #834, opened 2026-05-12
fix(frontend): align passport tabs with real backend contract + real world map

ErenCanOzkaya:
Passport detail UI shipped against a mock-data shape that diverged from the merged backend contract (`#583`). The Map tab crashed on the canonical payload, Stamps grouped everything under "Other", Timeline showed slug labels instead of the human-readable description the backend sends, and the world map itself was hand-drawn rectangles. This PR aligns every tab with the live API and replaces the hand-drawn map with a real one.

## Backend contract (already shipped, verified against `/api/users/<u>/passport/`)

```jsonc
{
  "stamps": [
    { "id", "culture", "category" /* lowercase: recipe|story|heritage|exploration|community */, "rarity", "earned_at", "source_recipe", "source_story" }
  ],
  "culture_summaries": [
    { "culture", "recipes_tried", "stories_saved", "interactions", "rarity" }
  ],
  "timeline": [
    { "id", "event_type", "description" /* human-readable */, "timestamp", "related_recipe", "related_story", "stamp_rarity" }
  ],
  "active_quests": [
    { "id", "name", "description", "category", "target_count", "reward_type", "reward_value",
      "is_event_quest", "event_start", "event_end", "progress", "completed_at", "reward_claimed" }
  ]
}
```

### Map tab — was crashing, now renders a real world map
- **Crash:** `PassportMap` read `c.name.toLowerCase()` but the backend ships `culture` (not `name`). Any non-empty `culture_summaries[]` payload threw `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` and dropped the tab into a white screen.
- **Visual:** the previous "world map" was 10 hand-drawn `<rect>` blobs over a light-blue background — Daglar's own comment said "Replace with a proper SVG asset once available". This PR replaces them with `react-simple-maps` driven by `world-atlas/countries-110m.json` (Natural Earth derived, ~130 KB topology), rendered via `geoEqualEarth` projection. Every country is a real polygon; ocean tint, white country borders, hover drop-shadow, and a non-pointer popover are kept.
- **Culture → country mapping:** `utils/passportCultureRegions.js` translates backend sub-region names (`Aegean`, `Black Sea`, `Mediterranean`, `Balkan`, `Nordic`, etc.) into the country names that culturally belong to them. Lookup is case-insensitive, trims whitespace, returns `[]` for unknown values so the map silently skips them rather than throwing. Duplicate claims resolve by engagement (`recipes_tried + stories_saved`) — first-occurrence wins on ties.
- New deps: **`react-simple-maps@^3` + `world-atlas@^2`**. React 19 peer is flagged by the package as 16/17/18 only but the wrapper just uses refs + useState; runs fine with React 19 in practice. `Dockerfile` swapped to `npm ci --legacy-peer-deps` so production build matches local install behaviour.

### Stamps tab — wrong category grouping
- `CATEGORIES = ['Recipe', 'Story', …]` was PascalCase but backend ships `category: 'recipe'` (lowercase). Strict `s.category === cat` matched nothing, so every stamp landed under the "Other" bucket.
- Fix: store both `{ key: 'recipe', label: 'Recipe' }` pairs, compare via `s.category.toLowerCase() === key`, render the canonical label. Unknown categories still fall through to "Other" so the data isn't silently dropped.

### Timeline tab — slug labels instead of descriptions
- The component rendered `event.event_type?.replace('_', ' ')` ("stamp earned", "recipe tried"). Backend already ships a human-readable `description` field ("Earned a bronze stamp for Black Sea", "Tried recipe: Black Sea Collard Green Sarma"). Now prefers `description`, falls back to the slug only when `description` is missing.

### Quests tab — wrong deadline field
- Component read `quest.deadline`; backend ships `event_end` on event-quests. Accept either so the deadline pill renders for real event quests, and mock-shape callers keep working.

### `passportMapColors.getRegionFill` — wrong field set
- Was reading `culture.stamp_rarity` / `culture.recipe_count` / `culture.story_count` (mock shape). Now reads canonical `culture.rarity` / `culture.recipes_tried` / `culture.stories_saved`, with the old keys as fallbacks. `heritage_count` isn't exposed per-culture by the backend (yet), so the "heritage contributed" tier collapses into the legendary-rarity branch — any culture deep enough to earn a legendary stamp gets the dark fill.

## Tests

- **`PassportMap.test.jsx`** (new) — 5 cases: crash safety on `null`/`undefined`/empty cultures, canonical backend shape (no `name`), and bad entries (`culture: null`, no `culture` key at all).
- **`passportCultureRegions.test.js`** (new) — 9 cases covering the lookup table (case-insensitive, trimmed, unknown), the country index builder (empty input, mapped countries, unknown skip, engagement tie-break, legacy mock-shape).
- **`StampGrid.test.jsx`** (existing, +2) — grouping by lowercase backend category, unknown-category fallback.
- **`passportMapColors.test.js`** (existing, +4) — canonical backend shape under a `describe('canonical backend shape')` block.
- **`PassportTimeline.test.jsx`** (existing, +2) — prefer `description`, fall back to slug.
- **`QuestList.test.jsx`** (existing, +1) — `event_end` as deadline source.

Local `CI=true npm test` → **648 passed** (was 621 before this PR).
Local `npm run build` → clean.
Docker `compose up -d --build web` → clean.

## Manual smoke (against the dev stack)

- `/users/mehmet` → Map tab no longer crashes; Turkey + Black Sea coast countries are tinted amber for the "recipe tried" stamp on Black Sea; hover shows the country name + "Black Sea" culture + rarity + counts.
- Stamps tab shows the bronze stamp under "Recipe" (was "Other").
- Timeline shows "Earned a bronze stamp for Black Sea" / "Tried recipe: Black Sea Collard Green Sarma" instead of the bare slug.

## Out of scope

- A backend-level expansion of the culture → country mapping (the per-region rows currently only cover the seed regions we ship). New entries can be added incrementally as more `Region` records land.
- Per-stamp progress bar inside `StampGrid` (would need `compute_rarity` exposed on the API).

### DIFF
--- a/app/frontend/.npmrc
+++ b/app/frontend/.npmrc
@@ -1,0 +1,1 @@
+legacy-peer-deps=true
--- a/app/frontend/Dockerfile
+++ b/app/frontend/Dockerfile
@@ -5,7 +5,10 @@ FROM node:20-alpine AS build
 WORKDIR /app
 
 COPY package*.json ./
-RUN npm ci
+# react-simple-maps@3 declares its peer as react 16/17/18 only, but runs
+# fine against react 19 (the wrapper only uses refs + useState). Use
+# --legacy-peer-deps so `npm ci` doesn't reject the resolved tree.
+RUN npm ci --legacy-peer-deps
 
 COPY . .
 
--- a/app/frontend/README.md
+++ b/app/frontend/README.md
@@ -2,6 +2,17 @@
 
 This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
 
+## Peer dependencies note
+
+The project uses React 19, but `react-simple-maps@3.0.0` (used by the passport
+world map) only declares peer compatibility up to React 18. Versions 3.1+ that
+support React 19 are not published yet, so we accept the unmet peer for now.
+
+To keep `npm install` / `npm ci` working without surprises, a local `.npmrc`
+sets `legacy-peer-deps=true`. The Docker build (`Dockerfile`) passes the same
+flag explicitly. If you ever bypass the `.npmrc` (e.g. running npm from a
+different cwd), use `npm install --legacy-peer-deps`.
+
 ## Available Scripts
 
 In the project directory, you can run:
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -20,7 +20,9 @@
         "react-leaflet": "^5.0.0",
         "react-router-dom": "^6.30.3",
         "react-scripts": "5.0.1",
-        "web-vitals": "^2.1.4"
+        "react-simple-maps": "^3.0.0",
+        "web-vitals": "^2.1.4",
+        "world-atlas": "^2.0.2"
       }
     },
     "node_modules/@adobe/css-tools": {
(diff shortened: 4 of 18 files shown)


mustafaocakxyz (approved):
LGTM as well

PR #836, opened 2026-05-12
fix(devops): seed the DB on backend start + fix web healthcheck (#835, #840)

laternis:
Closes #835
Closes #840

## Summary
**`app/backend/docker-entrypoint.sh` (#835):** The backend container only ran `migrate` + `collectstatic` on start, so production (and any fresh box brought up via `ops/PROD.md`) came up with an empty DB — 1 recipe, 0 stories, 0 cultural facts/content/events — and the live site rendered bare (`/api/cultural-facts/random/` 404, empty `/api/stories/`, `/api/recipes/`, `/api/map/regions/`). The `seed_*` management commands and their fixtures are shipped but were never invoked. Fixed by running the idempotent `seed_*` commands from the entrypoint, in dependency order: `seed_canonical` → `seed_region_geodata` → `seed_region_geo` → `seed_story_coordinates` → `seed_cultural_facts` → `seed_cultural_content` → `seed_ingredient_densities` → `seed_ingredient_routes`. A failing seeder logs a warning and the entrypoint continues; only `migrate` stays fatal. `seed_test_db` is intentionally excluded — it's a mock-data seeder for throwaway test DBs, not for prod.

**`docker-compose.yml` (#835):** Added `RUN_SEEDERS: ${RUN_SEEDERS:-1}` to the `backend` service so the entrypoint's seeding step is on by default for the compose stack (dev and prod) but off for the bare image. Bumped the `backend` healthcheck `start_period` 40s → 180s — gunicorn only `exec`s after the entrypoint finishes, so the first start against a fresh DB now does migrate + collectstatic + 8 seeders before the liveness probe can pass.

**`docker-compose.prod.yml` / `docker-compose.yml` (#840):** The `web` (nginx) container was permanently `(unhealthy)` even though it served traffic fine — `wget http://localhost/` can't pass with the prod nginx config: it only listens on `0.0.0.0` (the base image's `[::]:80` patch is skipped because `ops/nginx-prod.conf` is mounted read-only, so `localhost` → `::1` is refused), and the `:80` server block 301-redirects to https with the `genipe.app` cert (following the redirect fails verification). Overrode the `web` healthcheck in `docker-compose.prod.yml` to hit `https://127.0.0.1/` with `--no-check-certificate` (serves the SPA `index.html`, returns 200); the base `docker-compose.yml` keeps a plain-HTTP check for the dev/default image (no TLS listener there), now pinned to `127.0.0.1` instead of `localhost`.

**`.env.example`:** Documented `RUN_SEEDERS` (default `1`; set `0` to bring the stack up against an empty DB).

**`ops/PROD.md`:** Added a "Seeding" section (the command list + a manual `compose exec` snippet), added `RUN_SEEDERS` to the env table, updated the first-time-setup note and the `backend`/`web` healthcheck descriptions.

**`.github/workflows/deploy-web.yml`:** Updated the inline comment to reflect that the entrypoint now also runs the seeders.

All `seed_*` commands are idempotent (keyed on natural keys), so re-running on every container restart does not duplicate rows.

## Test plan
 - [ ] `sh -n app/backend/docker-entrypoint.sh` exits clean (syntax)
 - [ ] `docker compose -f docker-compose.yml config` and `docker compose -f docker-compose.yml -f docker-compose.prod.yml config` both parse with no error
 - [ ] Bring the prod stack up against an empty DB volume → `backend` reaches `(healthy)` within `start_period`; `recipes`, `stories`, `heritage_groups`, `cultural_facts`, `cultural_content`, `cultural_events` are all non-zero
 - [ ] `GET /api/cultural-facts/random/` returns 200; `/api/stories/`, `/api/recipes/`, `/api/map/regions/` are non-empty
 - [ ] `docker compose -f ... -f docker-compose.prod.yml ps` reports `web` as `(healthy)`
 - [ ] Restart the `backend` container → no duplicate rows (idempotency); entrypoint logs `running seeders`
 - [ ] `RUN_SEEDERS=0 docker compose up` → entrypoint logs `skipping seeders`, DB stays empty
 - [ ] Plain `docker compose up` (no prod overlay) → `web` healthcheck still passes (plain HTTP)
 - [ ] Existing dev/test workflows unaffected (`POSTGRES_HOST` unset → SQLite path; bare backend image has `RUN_SEEDERS` unset → no seeding)
 - [ ] `https://genipe.app/` still serves after deploy (regression)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,6 +19,11 @@ POSTGRES_PASSWORD=change-me
 POSTGRES_HOST=db
 POSTGRES_PORT=5432
 
+# Run the idempotent seed_* management commands on backend container start.
+# Defaults to 1 for the compose stack (see docker-compose.yml); set to 0 to
+# bring the stack up against an empty DB.
+RUN_SEEDERS=1
+
 # Frontend build-time (baked into the web image). Set this before
 # `docker compose build` — runtime changes are ignored.
 REACT_APP_API_URL=http://localhost
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -73,8 +73,10 @@ jobs:
               docker stop genipe-db || true
             fi
 
-            # Bring up the compose stack. The backend entrypoint runs migrate
-            # and collectstatic automatically before gunicorn starts.
+            # Bring up the compose stack. The backend entrypoint runs migrate,
+            # collectstatic, and the idempotent seed_* commands (RUN_SEEDERS,
+            # on by default) before gunicorn starts; that's covered by the
+            # backend healthcheck's 180s start_period.
             # --force-recreate guards against stale containers that survive a
             # config change (e.g. when the prod overlay adds ports or volumes
             # that compose's normal diff misses on already-running containers).
--- a/app/backend/docker-entrypoint.sh
+++ b/app/backend/docker-entrypoint.sh
@@ -7,6 +7,32 @@ python manage.py migrate --noinput
 echo "[entrypoint] running collectstatic..."
 python manage.py collectstatic --noinput
 
+# Seed reference + demo content. All seed_* commands are idempotent and
+# dependency-ordered (see each command's docstring), so this is safe to run on
+# every start. Gated by RUN_SEEDERS so the bare image stays generic — the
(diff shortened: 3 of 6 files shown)


roboticrustacean (approved):
lgtm

PR #846, opened 2026-05-12
fix(devops): move DB seeding into a separate manual workflow (#845)

laternis:
Closes #845

## Summary
`app/backend/docker-entrypoint.sh`: #836 made the entrypoint run the `seed_*` commands on every container start (gated by `RUN_SEEDERS`, on by default for the compose stack). In practice that took `genipe.app` fully down for ~1–3 min on **every merge to `main`** — `deploy-web.yml` does `docker compose up -d --force-recreate`, the entrypoint runs `migrate` + `collectstatic` + 8 seeders (including `seed_canonical`, which *wipes and rebuilds* ~84 recipes / 51 stories / users / media) before it `exec`s gunicorn, and `web` `depends_on: backend: service_healthy`, so nothing serves until seeding finishes. Re-seeding on every deploy is also wrong on its own: `seed_canonical._wipe()` destroys any account/recipe created on prod since the last deploy. Reverted the entrypoint to `migrate` + `collectstatic` only.

`docker-compose.yml` / `.env.example`: Removed the `RUN_SEEDERS` env plumbing; put the `backend` healthcheck `start_period` back to `40s` (no seed step to cover anymore).

`.github/workflows/seed-db.yml` (new): A `workflow_dispatch` action that runs the seeders against the running stack via `docker compose exec backend python manage.py …`. Inputs: `mode` = `safe` (everything except `seed_canonical` — all idempotent upserts) or `full` (includes the destructive `seed_canonical`; refuses to run unless `confirm=yes`). Shares `deploy-web.yml`'s `concurrency` group (`deploy-web-main`) so a seed run and a deploy can't overlap (a deploy recreates the backend container mid-seed). Prints DB counts at the end.

`.github/workflows/deploy-web.yml`: Updated the inline comment — the entrypoint no longer seeds; seeding is the separate workflow.

`ops/PROD.md`: Rewrote the "Seeding" section — seeding is a deliberate one-time / on-fixture-change operation, run via the new workflow (`mode=safe`/`full`) or `docker compose exec`; deploy first when fixtures change so the rebuilt image carries them. Added a `# 7. Seed the DB` step to first-time setup, removed the `RUN_SEEDERS` env-table row, and put the `backend` healthcheck description back to `40s`.

Note: prod is already seeded; the operator set `RUN_SEEDERS=0` in the server `.env` as an interim mitigation — once this merges, that var is unused and harmless. The `web` healthcheck fix from #840 is unrelated and untouched.

## Test plan
 - [ ] `sh -n app/backend/docker-entrypoint.sh` exits clean
 - [ ] `docker compose -f docker-compose.yml config` and `… -f docker-compose.prod.yml config` both parse
 - [ ] `.github/workflows/seed-db.yml` and `deploy-web.yml` parse as YAML
 - [ ] Merge to `main` → deploy recreates the backend container; `genipe.app` stays up (only the normal few-second nginx/gunicorn blip), no minutes-long outage; backend reaches `(healthy)` within `start_period`
 - [ ] Run the "Seed database" workflow with `mode=safe` → seeders run, DB counts printed, no data wiped
 - [ ] Run with `mode=full` and no `confirm` → workflow fails at the guard step
 - [ ] Run with `mode=full` + `confirm=yes` → `seed_canonical` + the rest run; counts non-zero
 - [ ] `grep RUN_SEEDERS docker-compose.yml .env.example` → nothing
 - [ ] `ops/PROD.md` Seeding section reads correctly

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,11 +19,6 @@ POSTGRES_PASSWORD=change-me
 POSTGRES_HOST=db
 POSTGRES_PORT=5432
 
-# Run the idempotent seed_* management commands on backend container start.
-# Defaults to 1 for the compose stack (see docker-compose.yml); set to 0 to
-# bring the stack up against an empty DB.
-RUN_SEEDERS=1
-
 # Frontend build-time (baked into the web image). Set this before
 # `docker compose build` — runtime changes are ignored.
 REACT_APP_API_URL=http://localhost
--- a/.github/workflows/deploy-web.yml
+++ b/.github/workflows/deploy-web.yml
@@ -73,10 +73,10 @@ jobs:
               docker stop genipe-db || true
             fi
 
-            # Bring up the compose stack. The backend entrypoint runs migrate,
-            # collectstatic, and the idempotent seed_* commands (RUN_SEEDERS,
-            # on by default) before gunicorn starts; that's covered by the
-            # backend healthcheck's 180s start_period.
+            # Bring up the compose stack. The backend entrypoint runs migrate
+            # and collectstatic automatically before gunicorn starts. Seeding
+            # is a separate manual step — see the "Seed database" workflow
+            # (.github/workflows/seed-db.yml) and ops/PROD.md.
             # --force-recreate guards against stale containers that survive a
             # config change (e.g. when the prod overlay adds ports or volumes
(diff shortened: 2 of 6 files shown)


roboticrustacean (approved):
thanks for the fix

PR #859, opened 2026-05-12
feat(frontend): inline action buttons and fix save-as-draft on recipe/story forms (#857)

Daglar1500:
## Summary

RecipeCreatePage: \"Save as draft\" and \"Publish Recipe\" buttons were stacked vertically on the left. Moved them inline (side by side), right-aligned, with a hint text below (\"Drafts stay private to you. Published recipes are visible to everyone.\"). Draft save validation relaxed to require only a title — description and ingredients are no longer required to save a partial draft.

StoryCreatePage: Same button layout applied for visual consistency. Added matching hint text (\"Drafts stay private to you. Published stories are visible to everyone.\").

Closes #857

## Test plan

 - [ ] `/recipes/new` — buttons appear side by side, right-aligned
 - [ ] `/recipes/new` — click \"Save as draft\" with only a title filled in → draft saves without validation errors
 - [ ] `/recipes/new` — fill all fields, click \"Publish Recipe\" → publishes normally
 - [ ] `/stories/new` — buttons appear side by side, right-aligned with hint text below
 - [ ] `/stories/new` — click \"Save as draft\" with title + body → draft saves
 - [ ] Both pages: hint text visible below buttons

### DIFF
--- a/app/frontend/src/__tests__/RecipeCreatePage.test.jsx
+++ b/app/frontend/src/__tests__/RecipeCreatePage.test.jsx
@@ -128,6 +128,30 @@ describe('RecipeCreatePage', () => {
     );
   });
 
+  it('saves draft with only a title — no description or ingredients required', async () => {
+    recipeService.createRecipe.mockResolvedValue({ id: 3 });
+    renderPage();
+    await waitFor(() => screen.getByLabelText(/title/i));
+    fireEvent.change(screen.getByLabelText(/title/i), { target: { value: 'Incomplete Draft' } });
+    fireEvent.click(screen.getByRole('button', { name: /save as draft/i }));
+    await waitFor(() =>
+      expect(screen.getByText(/draft saved/i)).toBeInTheDocument()
+    );
+    expect(recipeService.createRecipe).toHaveBeenCalledWith(
+      expect.objectContaining({ is_published: false, title: 'Incomplete Draft' })
+    );
+  });
+
+  it('draft save still rejects invalid ingredient amount', async () => {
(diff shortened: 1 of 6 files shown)


ErenCanOzkaya (approved):
LGTM — straightforward UX cleanup with the right test coverage.

## What I verified
- **Relaxed draft validation is well-scoped**: `validate(publish)` only enforces description/video and ingredients when `publish === true`. The positive-amount guard still runs in both paths, so users can't silently save a draft with `-5` as an amount — tested explicitly in "draft save still rejects invalid ingredient amount".
- **Test coverage matches the change**: the new "saves draft with only a title" case mocks `createRecipe` and asserts the payload carries `is_published: false` and the title, which is exactly the relaxed contract. Existing publish-path tests are untouched, so the publish flow still enforces the full requirements.
- **Button layout**: nested `.recipe-form-actions-buttons` / `.story-form-actions-buttons` row inside the existing right-aligned column gives buttons side-by-side with the hint underneath — clean and matches the screenshot in #857.
- **CI**: all three checks green (Backend / Frontend / Mobile).

## Minor (non-blocking)
- `StoryCreatePage` uses `className="publish-note"` but the class is only defined in `RecipeCreatePage.css`. Today it works because `App.js` eager-imports both pages, so both CSS files land in the main bundle. If anyone later moves to `React.lazy` for these routes, the story-page hint will silently lose its muted styling. Worth either duplicating the rule into `StoryCreatePage.css` or hoisting it to a shared stylesheet — but not a blocker for this PR.

Approving.

PR #862, opened 2026-05-12
feat(frontend): Cultural Highlights Read more → linked content or new detail page

ErenCanOzkaya:
Two-layer fix for the Explore page's **For You: Cultural Highlights** Read more button, which used to dump every title into \`/search?q=...\` and surface zero hits.

## What this PR does

1. **Use the backend's link when it has one.** \`CulturalContentCardSerializer\` already returns an optional \`link: { kind, id }\` per card pointing at the recipe / story / cultural event that owns the highlight. The frontend was dropping it.
2. **Show a real detail page when it doesn't.** Cards without a link now route to a new \`/highlights/:id\` page that renders the card's title, body, region and tags — instead of hiding Read more entirely.

## Changes

- \`culturalContentService.normalize\` passes \`link\` through, validating both \`kind\` and \`id\`; malformed payloads collapse to \`link: null\`.
- \`DailyCulturalSection.targetForItem\` resolves in priority order:
  - \`link.kind === 'recipe'\` → \`/recipes/:id\`
  - \`link.kind === 'story'\` → \`/stories/:id\`
  - \`link.kind === 'event'\` → \`/calendar\`
  - else → \`/highlights/<item.id>\`
  Read more is always shown.
- New page **\`CulturalHighlightPage\`** at \`/highlights/:id\` — re-fetches the daily highlights list and looks the id up (IDs are already stable strings like \`dc-fact-12\`). Renders title + body + region + tags. Not-found state links back to \`/explore\`; load-error state surfaces.
- New CSS for the highlight page (max-width 720, page-card style).

## Test plan
- [x] Full Jest suite: 726 passing (84 suites) · +8 net new tests
- [x] Production build clean
- [ ] Manual QA: open \`/explore\` → click a curated card. If the card is admin-linked to a recipe/story/event, it opens that. Otherwise it opens \`/highlights/...\` and shows the full body. Direct visit to \`/highlights/dc-fact-999\` shows the Not Found view.

## Follow-up
Admin/seed flows should consider populating \`link_kind\` + \`link_id\` for new \`CulturalContent\` rows when there's a natural target — the \`/highlights/:id\` page is a good fallback, but a direct recipe / story is better when one exists.

### DIFF
--- a/app/frontend/src/App.js
+++ b/app/frontend/src/App.js
@@ -30,6 +30,7 @@ import ModerationPage from './pages/ModerationPage';
 import HeritagePage from './pages/HeritagePage';
 import HeritageMapPage from './pages/HeritageMapPage';
 import CalendarPage from './pages/CalendarPage';
+import CulturalHighlightPage from './pages/CulturalHighlightPage';
 import NotFoundPage from './pages/NotFoundPage';
 
 export default function App() {
@@ -104,6 +105,7 @@ export default function App() {
           <Route path="/heritage/:id" element={<HeritagePage />} />
           <Route path="/heritage/:id/map" element={<HeritageMapPage />} />
           <Route path="/calendar" element={<CalendarPage />} />
+          <Route path="/highlights/:id" element={<CulturalHighlightPage />} />
           <Route path="*" element={<NotFoundPage />} />
         </Routes>
       </div>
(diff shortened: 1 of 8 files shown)


CaganIslam (approved):
All 726 tests pass, nice fallback page for unlinked highlights. LGTM

PR #864, opened 2026-05-12
feat(frontend): rework Explore — region rails, featured top, real images, skeletons (#863)

Daglar1500:
Summary
Explore page: The page was broken because `exploreService` filtered API results by `r.type` while `/api/recommendations/` returns `result_type` — every card fell through to a single generic "Discover" rail with empty placeholders. Normalized the payload, grouped by `region_tag`, surfaced the real `image` / `description` / `region_tag` / `rank_reason` fields that were already in the response, and reshaped the page around the regional-table framing.

Explore page: Added a "Today on Genipe" featured rail (top items by `rank_score`, taller cards), a region jump-nav chip bar, a sparse-rail grid layout (≤3 items), a per-rail author cap (max 2 per author), a 📖 glyph + `→ recipe` chip on story cards, a bookmark button on recipe cards for signed-in users, a hover overlay with description + `rank_reason`, skeleton rails during load, a sign-in banner for anonymous users, and an empty state with CTAs to /map and /recipes/new.

No backend changes.

Test plan
 [ ] Go to /explore — featured rail at top, then one rail per region (e.g. *Aegean*, *Marmara*, *Black Sea*); region badge is hidden inside region rails but visible on featured / catch-all
 [ ] Click a chip in the region jump-nav under the title — page scrolls to that rail
 [ ] A rail with 2 items renders as a left-aligned grid; a rail with 4+ items renders as a horizontal scroll
 [ ] Within a single rail, no author appears more than twice
 [ ] Each card shows the real cover image; force a broken `image` URL and confirm the gradient fallback shows with no console error
 [ ] Hover a card — description appears; if the item has a `rank_reason`, it shows under the description in italics
 [ ] Sign in, click the 🏷️ bookmark on a recipe card — icon flips to 🔖 (one network call to `/api/recipes/<id>/bookmark/`); click again — reverts
 [ ] Sign out — bookmark icon is gone; sign-in banner appears under the title
 [ ] Stub the API to return `[]` — empty state renders with "Open the map" and "Share a recipe" CTAs
 [ ] Stub the API to reject — error message renders
 [ ] During load, three skeleton rails are visible; no "Loading…" text; no layout jump when data arrives
 [ ] Story cards show 📖; stories with `linked_recipe_id` show a `→ recipe` chip that navigates to the linked recipe
 [ ] Open `/explore/region-aegean` (or whichever region exists) — EventDetailPage loads with the same items
 [ ] Run `CI=true npm test -- --testPathPattern="(ExplorePage|exploreService|EventDetailPage)"` — 19 tests pass

### DIFF
--- a/app/frontend/src/__tests__/ExplorePage.test.jsx
+++ b/app/frontend/src/__tests__/ExplorePage.test.jsx
@@ -2,39 +2,56 @@ import { render, screen, waitFor } from '@testing-library/react';
 import { MemoryRouter, Route, Routes } from 'react-router-dom';
 import ExplorePage from '../pages/ExplorePage';
 import * as exploreService from '../services/exploreService';
+import { AuthContext } from '../context/AuthContext';
 
 jest.mock('../services/exploreService');
+jest.mock('../services/recipeService', () => ({
+  toggleBookmark: jest.fn().mockResolvedValue({ is_bookmarked: true }),
+}));
 
 const mockEvents = [
   {
-    id: 'wedding',
-    name: 'Weddings',
-    emoji: '💍',
+    id: 'featured',
+    name: 'Today on Genipe',
+    emoji: '⭐',
+    featuredRail: true,
     featured: [
-      { type: 'recipe', id: 1, title: 'Wedding Soup', author_username: 'eren' },
-      { type: 'story', id: 2, title: "Our Big Day", author_username: 'alice' },
+      { type: 'recipe', id: 1, title: 'Wedding Soup', author_username: 'eren', region: 'Aegean' },
     ],
   },
   {
-    id: 'newyear',
-    name: 'New Year',
-    emoji: '🎉',
+    id: 'region-aegean',
+    name: 'Aegean',
+    emoji: '🫒',
     featured: [
-      { type: 'recipe', id: 3, title: 'Lentil Salad', author_username: 'bob' },
+      { type: 'recipe', id: 1, title: 'Wedding Soup', author_username: 'eren', region: 'Aegean' },
+      { type: 'story', id: 2, title: "Our Big Day", author_username: 'alice', region: 'Aegean' },
+    ],
+  },
+  {
+    id: 'region-marmara',
+    name: 'Marmara',
+    emoji: '🌊',
+    featured: [
+      { type: 'recipe', id: 3, title: 'Lentil Salad', author_username: 'bob', region: 'Marmara' },
     ],
   },
 ];
 
-function renderPage() {
+function renderPage({ user = { id: 1, username: 'me' } } = {}) {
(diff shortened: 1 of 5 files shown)


ErenCanOzkaya (approved):
LGTM — substantial rework, well-factored, ships much closer to the spec than the previous Explore.

## What I verified

- **Service layout is clean and testable**: `normalizeItem` / `capPerAuthor` / `pickFeatured` / `buildRails` are pure functions; the new tests cover (1) groups-by-region, (2) author cap within a rail, (3) type-fallback when no regions present, (4) empty results, (5) explicit `fetchEventDetail` lookup by region id, (6) backwards-compatible legacy `id=recipes` path. The `result_type` ↔ `type` normalization (the root cause noted in the PR body) is centralized in `normalizeItem`.
- **Auth-gated affordances are correctly gated**: sign-in banner shows only when `!user`, bookmark button only when `canBookmark && isRecipe`. `toggleBookmark` exists in `recipeService.js` (`POST /api/recipes/<id>/bookmark/`, idempotent — confirmed at `services/recipeService.js:115`).
- **Image fallback**: `onError` flips `imageBroken` → placeholder gradient takes over; `alt=""` is correct because the link's accessible name already carries the title.
- **Empty state + skeletons**: `hasContent` memo correctly checks all rails before falling through to the empty-state CTAs; skeleton rails are aria-hidden so SR users go straight to the error/empty handling.
- **Backwards compat**: `fetchEventDetail('recipes')` and `('stories')` still resolve via the new `buildRails` type-fallback path, so the legacy `/explore/recipes` URL keeps working.
- **CI**: all three checks green; `(ExplorePage|exploreService|EventDetailPage)` suites pass locally on my machine too (3 suites / 13 tests).

## Minor (non-blocking)

1. **Nested `<Link>` inside a `<Link>`** — the linked-recipe chip on story cards (`<Link to="/recipes/...">`) sits inside the outer `<Link to="/stories/...">`. That's invalid HTML (nested `<a>`) and depends on `stopPropagation()` + browser quirks to route the click correctly. Safer: replace the inner Link with a `<button type="button">` and call `navigate()` programmatically, or hoist the linked-recipe chip outside the outer Link.
2. **Bookmark state can desync**: `useState(false)` initializes the bookmark icon as unsaved regardless of the user's actual bookmark state, because the `/api/recommendations/` payload doesn't include `is_bookmarked`. First click is a no-op visually if the recipe was already bookmarked (the API call is idempotent so no data corruption, just a UX papercut). If the recommendation serializer ever surfaces `is_bookmarked`, threading it through `normalizeItem` would fix this.
3. **Dead ref**: `useRef(null)` in `RegionNav` is assigned but never read — can drop.

None of these block this PR. Approving.

PR #865, opened 2026-05-12
Mobile: passport map selection card, stamps layout, timeline embed

mustafaocakxyz:
Closes / relates to #860. 

Map: in-app selection card for pins; stamps 2-col compact; JourneyTimeline embedded + pagination sync; stats/level parity; region fit; etc.


CaganIslam (approved):
Passport map selection card, stamps layout and timeline embed all look good. 202/202 tests pass cleanly including the 4 new test files. LGTM

PR #866, opened 2026-05-12
feat: add User.avatar and User.display_name support #807

roboticrustacean:
Implemented backend support for user avatars (image upload) and human-readable display names. 
Updated models, serializers across all apps (users, recipes, stories, messaging), and added a dedicated upload endpoint at POST /api/users/me/avatar/.

closes #807


Emirhan777 (approved):
Lgtm.

PR #867, opened 2026-05-12
fix(mobile): contactability toggle reverts on save (#450)

CaganIslam:
## Summary
ContactabilityToggle: The toggle was visually reverting back to its previous value right after a successful save. Root cause was a stale-closure race: after the server PATCH returned, we spread the captured `user` from the closure on top of the server response (`await updateUser({ ...user, ...updated })`), which raced with the optimistic update that ran one line earlier and overwrote the new `is_contactable` with the older snapshot. Fix: use the server response directly (`await updateUser(updated)`) — it already contains the new flag plus any backend side-effects.

## Test plan
- [x] Log in → Profile → Messages → toggle OFF: switch stays OFF, label flips to "Block new threads"
- [x] Navigate Home → Messages again: toggle is still OFF (state persisted)
- [x] Toggle back ON: switch stays ON, label flips to "Allow new threads"
- [x] `npx tsc --noEmit` clean
- [x] `npx jest` 193/193 pass

### DIFF
--- a/app/mobile/src/components/ContactabilityToggle.tsx
+++ b/app/mobile/src/components/ContactabilityToggle.tsx
@@ -29,7 +29,11 @@ export function ContactabilityToggle() {
       // an unhandled promise rejection that crashes the toggle.
       await updateUser({ ...user, is_contactable: next });
       const updated = await updateMe({ is_contactable: next });
-      await updateUser({ ...user, ...updated });
+      // Use the server response as source of truth — it already contains the
+      // new `is_contactable` plus any side-effects (timestamps, etc). Spreading
+      // a stale `user` from the captured closure on top was racing with the
+      // optimistic update above and silently reverting the toggle (#450).
+      await updateUser(updated);
     } catch (e) {
       try {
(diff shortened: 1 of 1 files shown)


ErenCanOzkaya (approved):
LGTM — minimal, correct fix.

## What I verified
- `updateMe()` (`app/mobile/src/services/authService.ts:41`) returns `Promise<AuthUser>` — the full profile shape, not a partial — so `await updateUser(updated)` is safe and doesn't drop fields the rest of the app relies on.
- `updateUser` in `AuthContext` replaces both `setUser` state and the `AsyncStorage` cache with the argument, so the server response now flows end-to-end without being remixed with stale closure data.
- The previous `{ ...user, ...updated }` form depended on `updated` always carrying `is_contactable` to win the spread. Trusting the server response as source of truth removes that fragile dependency entirely (and matches the pattern used elsewhere after a server-confirmed PATCH).
- All three CI checks green; jest 193/193 and tsc clean per PR body.

Approving.

PR #868, opened 2026-05-12
fix(frontend): floating prompt scroll + cultural highlights always to detail page

ErenCanOzkaya:
Two adjacent fixes on the Explore page.

## 1. Floating cultural prompt overflow

The "What would you like to taste this week? ✨" floating prompt grew past the viewport when the user clicked **+50 more** on the region chip list — the overlay clipped behind the screen edge and there was no way to reach the chips at the bottom.

\`.floating-prompt\` now has \`max-height: calc(100vh - 4rem)\` + \`overflow-y: auto\`, so it scrolls inside its own box instead of clipping.

## 2. Cultural Highlights "Read more" landing on broken recipes

PR #862 routed Read more straight to \`/recipes/:id\` / \`/stories/:id\` when the backend gave a \`link\`. In practice the linked recipe / story has sometimes been removed since the highlight was authored, so the user lands on **"Could not load recipe"** with no way back to the highlight body.

Now:
- **\`DailyCulturalSection\`**: Read more always opens \`/highlights/:id\`. The detail page body is the guaranteed available content.
- **\`CulturalHighlightPage\`**: when the item carries a \`link\`, the page surfaces a secondary "View the linked recipe →" / "Read the linked story →" / "See on the cultural calendar →" action below the body. If that target is stale, the primary read still works — only the secondary link errors, which is recoverable.

## Test plan
- [x] Full Jest suite: 740 passing (84 suites)
- [x] Production build clean
- [ ] Manual QA: open Explore, expand the +50 more chip list — overlay scrolls. Click any highlight Read more — opens \`/highlights/...\` with full body; if linked content exists, the related-link button works.

### DIFF
--- a/app/frontend/src/__tests__/CulturalHighlightPage.test.jsx
+++ b/app/frontend/src/__tests__/CulturalHighlightPage.test.jsx
@@ -55,4 +55,32 @@ describe('CulturalHighlightPage', () => {
     renderAt('dc-fact-1');
     expect(await screen.findByText(/could not load this highlight/i)).toBeInTheDocument();
   });
+
+  it('surfaces a "View the linked recipe" link when item.link.kind=recipe', async () => {
+    culturalContentService.fetchDailyCulturalContent.mockResolvedValue([
+      { id: 'dc-fact-1', kind: 'fact', title: 'X', body: '', region: null, tags: [], link: { kind: 'recipe', id: 42 } },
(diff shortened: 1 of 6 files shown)


CaganIslam (approved):
Floating prompt scroll fix and cultural highlights fallback look solid. 740/740 tests pass. LGTM

PR #882, opened 2026-05-12
feat(frontend): SiteHeader user menu rework + Avatar component (#879)

Daglar1500:
Summary
SiteHeader: The user control was awkward — `A ▾` toggled the dropdown while `@ayse` was a separate link to the profile. Reordered to `A @ayse ▾` as a single toggle button. Removed the standalone username link.

Dropdown: "My Account" now navigates to `/users/<username>` (the user's own public profile) instead of `/account`. "Profile" still points at `/profile`.

New `Avatar` component (`app/frontend/src/components/Avatar.jsx`): renders `` when the field is present, falls back to initials on missing or broken image. Used in the Navbar and on `UserProfilePage`. Backend support for `profile_picture` is tracked in #878 — once that lands, photos appear automatically with no frontend follow-up.

UserProfilePage: Swapped the initials div for `<Avatar size="md">`. Capitalized the existing own-profile button label to "Edit Preferences".

AccountPage: Added a `← Back to my profile` link at the top to close the round-trip from `/users/:username`.

Tests: New `Avatar.test.jsx` (initials, ?, img, onError fallback, size class). Navbar test now also asserts "My Account → /users/:username" and that `@username` is inside the toggle button, not a separate link.

Test plan
 [ ] Logged-in header reads `A @ayse ▾`; clicking anywhere on it toggles the dropdown
 [ ] Dropdown → **My Account** navigates to `/users/<my-username>`
 [ ] Dropdown → **Profile** still navigates to `/profile`
 [ ] On `/users/<my-username>`, the "Edit Preferences" button is visible and links to `/account`
 [ ] On `/users/<someone-else>`, the "Edit Preferences" button is absent
 [ ] `/account` shows "← Back to my profile" linking back to `/users/<my-username>`; saving preferences still persists (PATCH /api/users/me/)
 [ ] Avatar shows initials everywhere (until #878 ships); when a user has `profile_picture`, the avatar renders the image; if the image URL is broken, it falls back to initials with no console error
 [ ] `CI=true npm test -- --testPathPattern="(Navbar|Avatar|AccountPage)"` — passes
 [ ] Logged-out header still shows Log In / Sign Up

### DIFF
--- a/app/frontend/src/__tests__/Avatar.test.jsx
+++ b/app/frontend/src/__tests__/Avatar.test.jsx
@@ -1,0 +1,34 @@
+import { render, screen, fireEvent } from '@testing-library/react';
+import Avatar from '../components/Avatar';
+
+describe('Avatar', () => {
+  it('renders the username initial when no profile_picture is set', () => {
+    render(<Avatar user={{ username: 'ayse' }} size="sm" />);
+    expect(screen.getByText('A')).toBeInTheDocument();
+  });
+
+  it('renders ? when there is no user', () => {
+    render(<Avatar user={null} size="sm" />);
+    expect(screen.getByText('?')).toBeInTheDocument();
+  });
+
+  it('renders an <img> when profile_picture is present', () => {
+    render(<Avatar user={{ username: 'ayse', profile_picture: '/media/avatars/a.jpg' }} size="md" />);
+    const img = document.querySelector('img.avatar-image');
+    expect(img).not.toBeNull();
+    expect(img).toHaveAttribute('src', '/media/avatars/a.jpg');
+    expect(screen.queryByText('A')).not.toBeInTheDocument();
+  });
+
+  it('falls back to initials when the image fails to load', () => {
+    render(<Avatar user={{ username: 'ayse', profile_picture: '/broken.jpg' }} size="sm" />);
+    const img = document.querySelector('img.avatar-image');
+    fireEvent.error(img);
+    expect(screen.getByText('A')).toBeInTheDocument();
+  });
+
+  it('applies the size variant class', () => {
+    const { container } = render(<Avatar user={{ username: 'a' }} size="md" />);
+    expect(container.querySelector('.avatar-md')).not.toBeNull();
+  });
+});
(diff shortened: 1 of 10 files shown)


ErenCanOzkaya (approved):
LGTM — clean UX rework with a tidy reusable Avatar component.

## What I verified

- **Navbar consolidation is correct**: the wrapper div + nested button-plus-link is collapsed into a single `<button>` with `aria-haspopup` / `aria-expanded` / `aria-label`. The `aria-label` (`User menu for @ayse`) takes precedence over the button's text content, so SRs announce the menu purpose cleanly while sighted users see `[A] @ayse ▾`. The Avatar's initial is correctly `aria-hidden`.
- **Avatar component is well-scoped**: simple API (`user`, `size`, `className`), `onError` flips to initials fallback, size variants via CSS classes only. Test cases cover initials / null-user / image-present / broken-image / size-class — that's exactly the surface to lock in.
- **Routing change is intentional and tested**: `My Account` → `/users/<username>` with a new Navbar test asserting the href. `Profile` still → `/profile`. The new `← Back to my profile` link on AccountPage closes the round-trip from `/users/:username` → `/account` → back.
- **Backwards compat**: until backend `profile_picture` (tracked in #878) ships, every site shows initials — no breakage. Once it lands, photos appear automatically with zero frontend follow-up, matching the pattern used elsewhere.
- **CI**: all three checks green; (Navbar|Avatar|AccountPage) suites pass per PR body.

## Minor (non-blocking)

- **Other initials renderers still exist**: `ChatTray.jsx:183`, `InboxPage.jsx:120`, `ThreadPage.jsx:96` all keep the inline `username?.[0]?.toUpperCase() ?? '?'` pattern. Replacing them with `<Avatar size="sm">` in a follow-up would (a) get them photo support for free when #878 lands and (b) avoid drift in fallback behavior (e.g., if someone tweaks the Avatar's `?` rule). Out of scope here.
- **Test plan checkbox count is a bit ambitious vs. the assertions** — only Navbar's "My Account → /users/:username" is directly asserted; the dropdown click-toggle, AccountPage back-link, and own-vs-other "Edit Preferences" visibility rely on existing coverage. Not blocking; the new pieces that needed coverage (Avatar fallbacks, routing change) got it.

Approving.

PR #884, opened 2026-05-12
feat(mobile): clean profile tab and add rich user info header (#873, #874)

CaganIslam:
Summary
ProfileTabScreen: Removed redundant first-layer rows (My recipes, My stories, Saved recipes, Cultural preferences, Edit profile) since UserProfile already exposes Recipes/Stories/Saved as tabs and Edit profile is accessible from there. Kept avatar/username card + Messages + Log out. Also fixed the Log out icon which was rendering as the literal string \u{1F6AA} because JSX attribute strings don't process escape sequences — switched to a raw 🚪 character. (#873)

UserProfileScreen: Added a rich header card matching the web UserProfilePage — region pill, join year, bio, "X recipes · Y stories" stats line, and chip rows for Cultural Interests / Dietary Preferences / Event Interests. Added an "Edit profile" button on own-profile that navigates to EditProfile, replacing the removed row in ProfileTabScreen. Non-own profiles continue to show the existing "Message" button. (#874)

profileService: Added fetchPublicProfile(username) hitting GET /api/users/<username>/ (note: not /profile/ — that path 404s). Comments document two other backend surprises: religious_preferences is the field name behind the UI's "Dietary Preferences" label, and created_at is what we use for the join date.

Test plan
 [ ] Profile tab shows only the avatar card, Messages, and Log out — no My recipes / My stories / Saved recipes / Cultural preferences rows
 [ ] Log out row icon renders as a door emoji, not the literal text \u{1F6AA}
 [ ] Open own profile from the avatar card — rich header (region, joined year, bio, stats, chip rows) renders
 [ ] Tap "Edit profile" on own profile — navigates to EditProfile screen
 [ ] Open another user's profile (e.g. via "By mehmet" chip on a story) — rich header renders, "Message" button shown, "Edit profile" button NOT shown
 [ ] Recipes / Stories / Saved tabs still work on both own and other-user profiles

### DIFF
--- a/app/mobile/src/screens/UserProfileScreen.tsx
+++ b/app/mobile/src/screens/UserProfileScreen.tsx
@@ -18,6 +18,10 @@ import {
   type BookmarkedRecipeListItem,
 } from '../services/bookmarkService';
 import { fetchPassport, type Passport } from '../services/passportService';
+import {
+  fetchPublicProfile,
+  type PublicUserProfile,
+} from '../services/profileService';
 import { fetchRecipesList } from '../services/recipeService';
 import { fetchStoriesList } from '../services/storyService';
 import { resolveTheme } from '../utils/passportTheme';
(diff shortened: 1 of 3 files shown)


mustafaocakxyz (approved):

Removing duplicate navigation to recipes/stories/saved/preferences/edit makes sense next to UserProfile’s tabs; keeping Messages + Log out is clear. Using a literal door emoji for the logout row fixes the broken escape-sequence icon.

Approve — LGTM.

PR #887, opened 2026-05-12
feat(frontend): calendar UX polish — single-month layout, smooth scroll, legend, themed detail (#886)

ErenCanOzkaya:
Closes #886. Four UX polish fixes on \`/calendar\`, one commit per concern.

## Changes

1. **Single-month layout when the month filter is engaged.** Picking a month from the dropdown now renders only that month's panel (the "Lunar / movable feasts" bucket still appears below if relevant). "All months" keeps the existing 12-panel grid.
2. **Smooth scroll to the detail panel on event click.** A \`useRef\` on the detail aside + a \`useEffect\` keyed on \`selected\` calls \`scrollIntoView({ behavior: 'smooth', block: 'center' })\` so the user doesn't have to scroll manually.
3. **Replaced the per-card lunar subline with a single legend.** The repetitive \`☾ On the lunar calendar: <Name> this year\` line that ran under every lunar card is gone. A two-swatch legend ("Mar 21 · Gregorian date", "Mar 19 · Lunar / movable feast") now sits below the calendar's intro paragraph.
4. **Detail panel tinted by event kind.** Selecting a Gregorian event gives the panel a light rust tint with a rust left border; resolved lunar events use a dark cream tint with a dark-brown border; unresolved/movable events use a mustard tint with a mustard border.

## Notes

- jsdom doesn't implement \`scrollIntoView\`. The component guards the call with \`typeof … === 'function'\` so the test environment is a no-op and real browsers scroll. The new "scrolls into view" test stubs the prototype method to assert it's called.
- Three of the \`#669\` companion tests asserted the per-card subline text; rewrote them to assert event presence + \`(movable)\` invariant instead, since the subline is intentionally removed by this PR.

## Test plan
- [x] Full Jest suite: **797 passing across 92 suites**
- [x] Production build clean (warnings only)
- [ ] Manual QA on \`/calendar\`:
  - Pick March from the month dropdown → only March + lunar bucket render.
  - Click any event card → page smooth-scrolls to the detail panel.
  - No \`On the lunar calendar\` line under any card; legend visible at the top.
  - Click a Gregorian event → rust-tinted panel. Click a lunar event → dark-cream panel. Pick a movable lunar slug → mustard panel.

### DIFF
--- a/app/frontend/src/__tests__/CalendarPage.test.jsx
+++ b/app/frontend/src/__tests__/CalendarPage.test.jsx
@@ -90,6 +90,15 @@ describe('CalendarPage', () => {
     expect(within(panel).getByRole('link', { name: /sumalak/i })).toHaveAttribute('href', '/recipes/10');
   });
 
+  it('scrolls the detail panel into view when an event card is clicked', async () => {
+    const scrollSpy = jest.fn();
+    window.HTMLElement.prototype.scrollIntoView = scrollSpy;
+    renderPage();
+    const card = await screen.findByRole('button', { name: /open nevruz details/i });
+    await userEvent.click(card);
+    expect(scrollSpy).toHaveBeenCalled();
+  });
+
   it('renders only the selected month panel when the month filter is set', async () => {
     culturalEventService.fetchCulturalEvents.mockResolvedValue([
       { id: 100, name: 'Spring Equinox', date_rule: 'fixed:03-21', region: { id: 1, name: 'Anatolia' }, description: '', recipes: [] },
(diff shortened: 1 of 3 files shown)


mustafaocakxyz (approved):

Single-month mode when the filter is set keeps the layout usable; lunar bucket still shows when needed. scrollIntoView behind a function check is the right approach for Jest/jsdom. Moving repetitive lunar copy into a shared legend and trimming cards improves scanability. tintForEvent + the three detail tint classes map cleanly to Gregorian vs resolved lunar vs movable. 

Approve — LGTM.

PR #888, opened 2026-05-12
feat(backend): expose recipe_title and story_title on PassportEventSerializer (#870)

laternis:
Summary
PassportEventSerializer: timeline events only sent the integer IDs of `related_recipe` / `related_story`, so mobile and web clients had to do an extra fetch per event to render a human-readable title for timeline pills. Added `recipe_title` and `story_title` as `SerializerMethodField`s that return the related object's `title` (or `null` when the FK is unset) — no extra API calls needed.

Closes #870.

Test plan
 [ ] `python manage.py test apps.passport` — all green
 [ ] GET `/api/users/<username>/passport/` — each `timeline` entry now has `recipe_title` and `story_title` keys
 [ ] A `recipe_tried` event shows `recipe_title` set and `story_title` null; a `story_saved` event shows the reverse
 [ ] An event with no related recipe/story (e.g. `level_up`) has both titles null


roboticrustacean (approved):
lgtm

PR #889, opened 2026-05-12
fix(frontend): cover all seeded regions in home map country lookup

ErenCanOzkaya:
## Summary

The home page's "Explore by region" world map was tinting only ~10 countries even though `/api/map/regions/` already returns site-wide content counts across 58 seeded culinary regions. The cause was entirely on the frontend: `utils/countryRegions.js` only mapped 20 countries — all Turkey-centric — so every other country fell through to the neutral cream branch with no hover popover and no click target. That made the map read as "this is just my data" even though the data was always global.

This PR keeps the visual identical (same rust gradient, same hover popover, same `/search?region=<name>` click destination) and only expands the country→region lookup so the rest of the world responds to data.

### Changes

- **`utils/countryRegions.js`** — expand `COUNTRY_TO_REGION` to ~140 entries covering every country/alias on the world-atlas map, mapped to the most specific backend `Region.name` from `app/backend/apps/recipes/migrations/0004_seed_regions.py`. Prefer country-level regions (`Russia` → `Russia`, `Spain` → `Spain`, `Brazil` → `Brazil`) over macro-regions (`Sweden` → `Nordic`, `Croatia` → `Balkan`, `Vietnam` → `Vietnam`).
- **Bug fix**: `Iraq` was mapped to `Southeast Anatolia` but the backend seeds `Southeastern Anatolia` (no `-ern` was a typo). Iraq used to silently stay cream even though the backend has data for it.
- **Aliases**: world-atlas / Natural Earth truncates a few long names with `.` (`Bosnia and Herz.`, `Czech Rep.`, `Dem. Rep. Congo`, `S. Sudan`). Both forms now resolve.
- **`__tests__/countryRegions.test.js`** (new) — 9 cases covering case-insensitive lookup, null on unmapped / non-string input, country-specific-over-macro precedence, macro fallback, Iraq regression, world-atlas alias coverage, and a contract test that every value in the mapping exists in the backend seed list.
- **`__tests__/RegionContentMap.test.jsx`** — fixture used `Anatolia` and `Aegean` for Turkey/Greece, but neither is a seeded region (`Anatolian` and `Greece` are). The test passed only because the old mapping had the same typo. Updated the fixture to use real region names so the test now exercises the actual contract.

### No backend changes

`/api/map/regions/` already returns site-wide counts and is `permissions.AllowAny` (no user filter — anyone gets the same answer). The map was always pulling the right data; the frontend just wasn't translating most countries into something the data could be joined against.

## Test plan

- [ ] Open `/` on a wide window — country tints visible across Europe, Asia, the Americas, Africa, and Oceania (not just Turkey + Mediterranean basin).
- [ ] Hover Japan — popover shows `Japanese · N recipes · M stories`. Hover Brazil — popover shows `Brazil · N recipes · M stories`. Hover Iraq — popover shows `Southeastern Anatolia · …` (the previously broken case).
- [ ] Click Japan — navigates to `/search?region=Japanese`.
- [ ] Hover Antarctica (or any country without a culinary-region mapping) — popover says "Not part of a culinary region yet", click does nothing.
- [ ] `cd app/frontend && CI=true npm test -- --watchAll=false --testPathPattern="(countryRegions|RegionContentMap)"` — 15 tests pass.
- [ ] `cd app/frontend && CI=true npm test -- --watchAll=false` — full suite green (93 suites / 806 tests).

## Out of scope

- Backend seed expansion. If a region the frontend lookup references is later renamed or removed, the new contract test (`every mapped region value matches a backend seed Region.name`) fails loudly so the drift is caught at PR time.
- The passport map's `passportCultureRegions.js` — that's a different flow (culture-name → countries) and keeps the engagement-weighted multi-culture-per-country behavior the passport needs.

### DIFF
--- a/app/frontend/src/__tests__/RegionContentMap.test.jsx
+++ b/app/frontend/src/__tests__/RegionContentMap.test.jsx
@@ -38,17 +38,18 @@ function renderMap(regions) {
   );
 }
 
-// Turkey → "Anatolia", Greece → "Aegean" per countryRegions.js.
+// Turkey → "Anatolian", Greece → "Greece" per countryRegions.js (names match
+// the backend seed in app/backend/apps/recipes/migrations/0004_seed_regions.py).
 const REGIONS = [
-  { id: 1, name: 'Anatolia', content_count: { recipes: 4,  stories: 2 } },
-  { id: 2, name: 'Aegean',   content_count: { recipes: 12, stories: 5 } },
+  { id: 1, name: 'Anatolian', content_count: { recipes: 4,  stories: 2 } },
+  { id: 2, name: 'Greece',    content_count: { recipes: 12, stories: 5 } },
 ];
 
 describe('RegionContentMap', () => {
   it('shows N recipes · M stories on hover for a mapped country', () => {
     renderMap(REGIONS);
     fireEvent.mouseEnter(screen.getByTestId('geo-Turkey'));
-    expect(screen.getByText(/anatolia/i)).toBeInTheDocument();
+    expect(screen.getByText(/anatolian/i)).toBeInTheDocument();
     expect(screen.getByText(/4 recipes.*2 stories/i)).toBeInTheDocument();
   });
 
--- a/app/frontend/src/__tests__/countryRegions.test.js
+++ b/app/frontend/src/__tests__/countryRegions.test.js
@@ -1,0 +1,100 @@
+import { COUNTRY_TO_REGION, getRegionForCountry } from '../utils/countryRegions';
+
+/**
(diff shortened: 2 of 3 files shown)


mustafaocakxyz (approved):

Expanding countryRegions so almost all atlas country names resolve to a seeded Region.name is the right fix for the “only Turkey lights up” issue; 

Iraq → Southeastern Anatolia matches 0004_seed_regions.py. The new countryRegions tests (including the orphan contract against the migration mirror) and the RegionContentMap fixture update to Anatolian / Greece give good regression coverage.

Approve — LGTM.

PR #891, opened 2026-05-12
feat(frontend): compact search forms on /recipes and /stories

ErenCanOzkaya:
\`/recipes\` and \`/stories\` had no way to search from inside themselves — the user had to jump to the explicit search page first. Added a small inline form above the grid on both pages. Submitting routes to \`/search?q=<query>\` which already handles cross-type results.

## Changes
- **RecipeListPage**: \`role=\"search\"\` form with input + Search button at the top of the page; aria-label \"Search recipes\".
- **StoryListPage**: same shape, aria-label \"Search stories\".
- Per-page CSS (\`.list-search-form\` / \`.list-search-input\` / \`.list-search-btn\`), capped at 560px so the form sits comfortably without dominating the grid.

## Note on the second ask (single-toggle pills)
The original request also included consolidating SearchPage's \`+ / -\` chip pairs into a single toggle pill. Implementation was started but rolled back during the session — that piece will land separately so this PR stays narrow.

## Test plan
- [x] +4 tests (2 per page): form renders, submit navigates
- [x] Full Jest suite green locally
- [ ] Manual QA: open \`/recipes\` → type in the search box → submit → land on \`/search?q=…\`. Repeat on \`/stories\`.

### DIFF
--- a/app/frontend/src/__tests__/RecipeListPage.test.jsx
+++ b/app/frontend/src/__tests__/RecipeListPage.test.jsx
@@ -143,3 +143,36 @@ describe('RecipeListPage bookmark icon', () => {
     expect(screen.getByLabelText(/not bookmarked/i)).toBeInTheDocument();
   });
 });
+
+describe('RecipeListPage — search form', () => {
+  beforeEach(() => {
+    recipeService.fetchRecipes.mockResolvedValue([]);
+  });
+
+  function renderWithRoutes() {
+    return render(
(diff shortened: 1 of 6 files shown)


mustafaocakxyz (approved):

The inline role="search" forms on /recipes and /stories reuse /search?q=…, which matches how the rest of the app centralizes discovery. Styling is restrained (max-width 560px, clear focus treatment), and the new tests cover render + navigation. 

Approve — lgtm.

Issue #95, opened 2026-03-05
Create the sequence diagram for the scenario of group member 2


Daglar1500:
## Description
This issue covers the creation of the Sequence Diagram for the scenario assigned to Group Member 2: **[Insert Scenario Name Here, e.g., Share a Story and Link an Existing Recipe]**. This diagram will model the dynamic interactions and method calls between system objects (such as `User`, `PlatformController`, `Story`, `Recipe`, etc.) over time for this specific flow.

## Motivation / Context
Having completed our static Class Diagram, we now need to clarify the dynamic behaviors and workflows of the system. We are creating sequence diagrams for approximately 9 different scenarios across the team. This issue is a crucial part of our system architecture documentation and will directly guide both our frontend and backend development phases.

## Scope
* Mapping the end-to-end interaction flow for the specific scenario assigned to Group Member 2.
* Utilizing the exact class names (e.g., `Content`, `PlatformController`) and methods (e.g., `shareStory()`, `linkRecipe()`) defined in our unified Class Diagram.
* Formatting the sequence diagram using Mermaid.js syntax so it renders correctly on GitHub.

## Deliverables
* A brief text description explaining the steps of the scenario.
* The fully functional Sequence Diagram code enclosed within ````mermaid ```` blocks.

## Acceptance Criteria
- [ ] Objectives are clearly defined
- [ ] Deliverables are produced and accessible
- [ ] Scope constraints are respected
- [ ] Work adheres to relevant guidelines or standards
- [ ] Outcome is reviewable and verifiable

Issue #12, opened 2026-02-11
Create the skeleton of wiki page


mustafaocakxyz:
Ufuk is requested to create the initial form of the wiki page in order for members to create their personal bios.

@ufukaltunbulak is requested to achieve the task.

@mustafaocakxyz is the reviewer to close the issue.

ufukaltunbulak:
Thank you @mustafaocakxyz for finishing before I do. It has a nice design.

PR #892, opened 2026-05-12
fix(mobile): show recipe/story titles in passport timeline pills (#871)

CaganIslam:
Closes #871

Summary
JourneyTimeline pills used to render the placeholder text \`Recipe #95 →\` / \`Story #12 →\` because the wire payload never carried a human-readable title. Backend #870 added \`recipe_title\` and \`story_title\` as SerializerMethodField on PassportEventSerializer, so the title is now available on every event that has a related recipe or story.

passportTimelineService.ts: Folded the two new backend fields into the event payload alongside \`related_recipe\` / \`related_story\` so the rendering layer keeps a single shape to read from. Empty strings are skipped so the existing ID fallback still kicks in for events the backend can't title (e.g. a recipe that was deleted).

JourneyTimeline.tsx: Pill text is now \`\${payload.recipe_title ?? 'Recipe #' + id} →\` (and the equivalent for stories). The accessibility label also reads the title when available, which is the bigger win for screen reader users — "Open recipe Trabzon Butter Kuymak" is a lot more useful than "Open Recipe 95". The hardcoded ID still ships as a fallback because older event rows in the DB will not have a backfilled title and we don't want to render nothing.

Tests
- 2 new JourneyTimeline tests: pill renders title when payload has recipe_title / story_title, and the old "Recipe #ID" string is no longer present
- 2 new passportTimelineService tests: normalizer folds titles into payload, and skips empty strings so the ID fallback still kicks in
- 35/35 timeline + service tests pass

Test plan
 [ ] Pull #870 backend changes locally, open Cultural Passport → Timeline as a user with recipe_tried / story_saved / stamp_earned events
 [ ] Pills now read the actual recipe/story title (e.g. "Trabzon Butter Kuymak →") instead of "Recipe #95 →"
 [ ] Tap a pill — still navigates to RecipeDetail / StoryDetail by ID, no regression
 [ ] Force an event without a title (e.g. recipe deleted) — pill falls back to "Recipe #ID →" instead of breaking
 [ ] Screen reader announces the title in the pill's accessibility label

### DIFF
--- a/app/mobile/__tests__/components/JourneyTimeline.test.tsx
+++ b/app/mobile/__tests__/components/JourneyTimeline.test.tsx
@@ -239,6 +239,34 @@ describe('JourneyTimeline', () => {
     expect(queryByText(/Story #/)).toBeNull();
   });
 
+  it('renders Recipe pill with title from payload when backend provides recipe_title (#871)', async () => {
+    const ev = makeEvent({
+      id: 30,
+      message: 'Saved a recipe',
+      payload: { related_recipe: 99, recipe_title: 'Trabzon Butter Kuymak' },
+    });
+    mockPaginationSync([ev]);
+    const { getByText, queryByText } = render(
+      <JourneyTimeline username="ayse" initialEvents={[ev]} />,
+    );
+    expect(getByText('Trabzon Butter Kuymak →')).toBeTruthy();
+    expect(queryByText('Recipe #99 →')).toBeNull();
+  });
+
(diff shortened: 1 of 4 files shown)


mustafaocakxyz (approved):

Folding recipe_title / story_title into payload in normalizeEvent (skipping empty strings) keeps the wire shape tidy. JourneyTimeline pills and accessibilityLabels now prefer human titles with a safe ID fallback; navigation still keys off related_* ids — no regression risk for older payloads. Tests cover both the normalizer and the pill copy.

Approve — LGTM.

PR #893, opened 2026-05-12
feat: implement backend feedback API endpoint #877

### LINKED ISSUE #877: Backend: feedback endpoint for the home page Feedback Bar (POST /api/feedback/)
## Description

Web home page redesign (#876) ships a thin **Feedback Bar** under the weekly rails: a single-line input where visitors can say "what would you like to see on Genipe?". For that PR the submission is a frontend-only stub. This issue is for the matching backend endpoint that replaces the stub with a real network call.

## Wanted

A small Django app (or addition to an existing one) exposing:

\`\`\`
POST /api/feedback/
\`\`\`

### Request

\`\`\`json
{
  "message": "I'd love to see more Andean recipes."
}
\`\`\`

- \`message\`: required, plain text, **2 ≤ length ≤ 2000** chars after trim.
- Both authenticated and anonymous submissions allowed. If the request is authenticated, the user FK is recorded; otherwise \`user\` is \`null\`.

### Response

- \`201 Created\` on success, body: \`{ "id": <int>, "created_at": <iso8601> }\` (no echo of the message — keeps payloads small).
- \`400 Bad Request\` with \`{ "message": ["This field may not be blank."] }\` (or the right DRF error shape) on validation failure.

### Throttling

- Anonymous: 5 / hour / IP (use DRF \`ScopedRateThrottle\` or similar) to keep the form from being abused as a complaint funnel.
- Authenticated: 30 / hour / user.

### Model

\`\`\`python
class Feedback(models.Model):
    user = models.ForeignKey(User, null=True, blank=True, on_delete=models.SET_NULL,
                             related_name='feedback_submissions')
    message = models.TextField()
    created_at = models.DateTimeField(auto_now_add=True)
    # Light optional metadata so the team can triage in admin:
    user_agent = models.CharField(max_length=255, blank=True, default='')
    path = models.CharField(max_length=255, blank=True, default='')

    class Meta:
        ordering = ['-created_at']
\`\`\`

\`user_agent\` and \`path\` are populated server-side from the request headers; they are **not** part of the request body.

### Admin

Register \`Feedback\` in \`admin.py\` with \`list_display = ('id', 'short_message', 'user', 'created_at')\` and \`readonly_fields = ('user', 'message', 'created_at', 'user_agent', 'path')\` so the team can read submissions without editing them.

## Acceptance Criteria

- [ ] \`POST /api/feedback/\` accepts a non-empty \`message\` and returns 201 with \`{ id, created_at }\`.
- [ ] Empty / whitespace-only / over-2000-char messages return 400 with a DRF error shape.
- [ ] Anonymous calls succeed (no auth required).
- [ ] Authenticated calls attach \`request.user\` to the saved row.
- [ ] Throttling: 5 / hour / IP for anon, 30 / hour / user.
- [ ] Admin lists feedback rows with the user, a truncated message preview, and \`created_at\`.
- [ ] Unit tests cover: happy path (anon), happy path (auth), blank rejection, length cap, throttle (mocked).

## Out of Scope

- A web admin moderation UI beyond Django admin.
- Email / Slack notifications on new feedback — file separately if wanted.
- Public read-back of feedback — keep it write-only from the API.

## References

- Web frontend issue: #876 (\`FeedbackBar\` component there will switch from \`feedbackService\` stub to a real \`apiClient.post('/api/feedback/', { message })\` once this lands).
- Frontend service file to point at after this is built: \`app/frontend/src/services/feedbackService.js\` (created in the #876 PR).

roboticrustacean:
This PR implements the `POST /api/feedback/` endpoint to handle user feedback from the home page Feedback Bar.

**Key changes:**
- Created `feedback` app with `Feedback` model and associated migration.
- Implemented `FeedbackSerializer` with 2-2000 char message validation and minimized response payload.
- Implemented `FeedbackCreateView` with custom throttling (5/hr anon, 30/hr auth).
- Configured metadata collection (`user_agent`, `path`) from request headers.
- Registered the model in Django admin with read-only access for triage.
- Added 8 unit tests covering happy paths, validation, and throttling.

closes #877


laternis (approved):
LGTM

PR #894, opened 2026-05-12
feat(frontend): SearchPage filter chips as single toggle pills

ErenCanOzkaya:
Each tag in **Dietary Tags**, **Event Tags**, and **Ingredients** on \`/search\` used to render as two pills side by side — \`+ Vegan\` and \`- Vegan\` — one for include and one for exclude. Visually noisy, and the exclude path was rarely used.

## Changes
- Each tag is now a **single pill** that toggles on click:
  - Default: outlined cream pill with the tag name.
  - Selected: filled rust pill, \`aria-pressed=\"true\"\`.
  - Click again → unselect.
- Removed the exclude state entirely: state variables, URL query params (\`ingredient_exclude\` / \`diet_exclude\` / \`event_exclude\`), the active-filter chip labels, and the unused \`setExcludeOnly\` helper.
- The active-filter chip below the filters now reads \"Ingredient: yogurt\" instead of \"Ingredient+: yogurt\".
- Search service params dropped the \`*_exclude\` keys; the existing include-only flow covers the common case.

## Test plan
- [x] +2 SearchPage tests cover toggle on / off and \`aria-pressed\` state
- [x] Existing 14 SearchPage tests updated for the new label / param shape
- [x] Full Jest suite: 807 passing
- [ ] Manual QA on \`/search\` → expand Dietary Tags → click \"Vegan\" → pill fills; click again → pill empties; submit → URL carries \`diet=Vegan\` only.

## Follow-up
Re-introduce a separate \"exclude\" affordance only if real users ask for it. The backend still accepts \`*_exclude\` params, so nothing breaks server-side.

### DIFF
--- a/app/frontend/src/__tests__/SearchPage.test.jsx
+++ b/app/frontend/src/__tests__/SearchPage.test.jsx
@@ -67,11 +67,8 @@ describe('SearchPage', () => {
     await waitFor(() => {
       expect(searchService.search).toHaveBeenCalledWith('soup', 'Aegean', '', {
         ingredient: 'Tomato',
-        ingredient_exclude: '',
         diet: '',
-        diet_exclude: '',
         event: '',
-        event_exclude: '',
         meal_type: '',
       });
     });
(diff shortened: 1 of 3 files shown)


laternis (approved):
LGTM

PR #895, opened 2026-05-12
Mobile: align onboarding with seed_canonical (#883)

mustafaocakxyz:
Aligns `OnboardingScreen` chip labels with `app/backend/fixtures/seed_canonical.json` so saved preferences match seeded data.
Closes #883


CaganIslam (approved):
Onboarding options aligned with seed_canonical. 206/206 tests pass. LGTM

PR #896, opened 2026-05-12
feat(mobile): color polish and linked recipe image

CaganIslam:
Summary
Tokens: bumped accentGreen from the deep olive #4A7722 to a softer #95c26e and surface from #E69138 to #e8a45a. The new green reads better against the warm cream background, and the lighter surface gives recipe and story cards more contrast against the rust app background. Also nudged the web --color-accent-green-text from #2D6A4F to #7fd496 for consistency with the mobile token (note: the new shade no longer hits WCAG AA on cream — flagged below).

ProfileTabScreen: Replaced the cream backgrounds on the @username header card and the Messages row with the pink #e09faa we already use for Log out. The screen now reads as a single coherent block of warm tones instead of two cream slabs floating in a rust card. Log out row hex literal stays the same — just lifted the same shade up to the other rows for visual rhyme.

LinkedRecipePreviewCard: Constrained the thumb to a fixed 96x96 square centred vertically. Before this the thumb had a fixed width but no fixed height, so as soon as we started filling it with a portrait recipe photo the image stretched and pulled the whole card to ~400px tall. Square thumb keeps the card compact regardless of the source image's aspect ratio.

StoryDetailScreen / storyService / story type: Backend's story serializer only exposes linked_recipe as an integer ID, so the LinkedRecipePreviewCard had nothing but a "R" letter to render. Wired the screen up to fetch the recipe by id after the story loads and pipe the returned image url into the preview card. One extra request per story view — accepted cost so we get a real thumbnail. Added image to the linked_recipe type and normalizer too, so the moment backend ships an embedded image field the cached call goes away naturally without another diff.

Caveats
- The web global.css comment still claims --color-accent-green-text is 5.97:1 AA-compliant on cream; with #7fd496 it's closer to 1.7:1. Either we walk this one back to the older darker green, or we drop the AA comment. Left as-is for now since the rest of the mobile palette work was the focus.

Test plan
 [ ] Open the Profile tab — @username card, Messages row and Log out row should all share the same pink background, no cream rectangles
 [ ] Browse Home — story and recipe cards use the new softer orange and the "From your region" pill uses the softer green
 [ ] Open a story that has a linked recipe — the Linked Recipe card shows the actual recipe photo in a tidy 96x96 square thumb, no oversized stretched image, no fallback "R" letter
 [ ] Open a story whose linked recipe has no image — fallback "R" still renders cleanly inside the same 96x96 thumb


mustafaocakxyz (approved):

Token tweaks and Profile row coloring read as a coherent polish pass. Linked recipe flow is sensible: 96×96 thumb fixes layout, and fetchRecipeById fills the gap when the story API only exposes an id; normalizeStoryDetail + StoryDetail type already prepare for an embedded image later.

Approve — LGTM.

PR #901, opened 2026-06-02
fix(mobile,backend): post-MVP photo loading, map crash and profile scroll fixes

### LINKED ISSUE #1: Add the group members into the the repo
The group members that are announced at the Moodle page for the group 12 of CMPE354 Software Engineering course in 2026 Spring term. The group members will be able to works on code, issues, pull requests, wiki etc. with the role they need to do those task to develop the project.

### LINKED ISSUE #2: Set the first introductory meeting before the first PS hour
All 10 group members of the project need to meet each other before the first PS hour of the course that is at 1pm on 11th Feb, 2026 in order to set the methods to collaborate successfully  and effectively.

### LINKED ISSUE #3: Create Wiki of the project
After getting the admin roles of the project, Wiki section of the project needs to be created with the necessary initial pages.

### LINKED ISSUE #6: Create personal pages to introduce members and track what works are done
Every group members are needed to create and regularly update his personal pages including information about themselves and trackable work performance by themselves.

CaganIslam:
## Note on preparation & disclosure

Most of these were small one-or-two-line issues that only surfaced in the standalone release build. When I run the mobile app locally, Expo automatically supplies the Google Maps API key, so the map crash never appeared on my machine; the local/dev build also auto-allows the http image traffic, so photos loaded fine there. Only the release APK was missing the key and blocked cleartext — so we added both explicitly. It may well work fine when you run it locally from the container too, but for honesty we wanted to flag this and provide the corrected version.

## Summary

Post-MVP testing on the standalone release APK surfaced six issues across the backend and the mobile client. Each is documented below with its cause, the change, and the resulting behaviour.

### 1. Recommendation cards had no photo (backend)

1. **Cause:** The `/api/recommendations/` endpoint returned the image as a host-less relative path (`/media/...`). Browsers auto-resolve such paths, but React Native's `<Image>` cannot, so it silently rendered nothing.
2. **What we changed:** In `apps/search/views.py`, the recipe and story payload helpers now return an absolute URL instead of the bare `image.url`.
3. **Result:** Home → "Recommended for you", "More recipes" and the Stories rail show their thumbnails instead of empty coloured cards.

### 2. Linked recipe showed an "R" placeholder (backend)

1. **Cause:** The story payload exposed the linked recipe's id and title but not its image, so clients had nothing to render and fell back to a letter placeholder.
2. **What we changed:** Added a `recipe_image` field (absolute URL via DRF `ImageField`) to `StoryRecipeLinkSerializer` and included it in `Meta.fields`. Purely additive — existing keys are untouched, so web and existing contract tests are unaffected.
3. **Result:** The story payload now carries the linked recipe's image; the "Linked Recipe" card can display the real photo.

### 3. Linked recipe image not surfaced on mobile (mobile)

1. **Cause:** The mobile story normalizer read the linked-recipe image from the legacy singular `linked_recipe` value, which the backend returns as a bare id — so no image ever reached the card.
2. **What we changed:** `storyService.ts` now reads `linked_recipes[0].recipe_image` (the array field populated by fix #2) and maps it onto `linked_recipe.image`, falling back to the legacy value when absent. The `StoryDetail` type already permits an optional `image`, and the preview card already renders one when present, so no type or component change was needed. (Depends on fix #2 being deployed.)
3. **Result:** The "Linked Recipe" card on Story Detail shows the recipe photo; recipes without a photo still fall back to "R" (no regression).

### 4. Profile screen could not scroll on the Map tab (mobile)

1. **Cause:** The outer `ScrollView` was conditionally disabled while the Passport "Map" tab was active (`scrollEnabled={activePassportTab !== 'map'}`), intended to protect map marker taps on Android. The side effect was that the whole screen became un-scrollable and the map was clipped at the bottom.
2. **What we changed:** Removed the condition in `UserProfileScreen.tsx`; `nestedScrollEnabled` alone now disambiguates the inner map gestures from the outer scroll.
3. **Result:** The profile page scrolls on the Map tab and the map's bottom edge is reachable, while region markers remain tappable.

### 5. Map screens crashed on the release APK (mobile)

1. **Cause:** `react-native-maps` requires the Google Maps API key as manifest meta-data. The standalone release APK did not declare it (Expo Go embeds its own key, which is lost once the app is standalone), so the first map view threw `IllegalStateException: API key not found` and the app crashed.
2. **What we changed:** Added the key under `expo.android.config.googleMaps.apiKey` in `app.json`, so Expo prebuild injects the required `com.google.android.geo.API_KEY` meta-data into the built manifest.
3. **Result:** Map Discovery, Heritage Map and Region Map Detail render without crashing.

### 6. Images blank on the release APK — cleartext traffic (mobile)

1. **Cause:** The backend serves media over `http://`. Android 9+ release builds block cleartext (non-TLS) traffic by default, so `<Image>` silently failed to load anything served over `http://`. A manual `AndroidManifest.xml` edit does not survive a build, because `android/` is generated by prebuild and gitignored.
2. **What we changed:** Added the `expo-build-properties` plugin in `app.json` with `android.usesCleartextTraffic: true`, so the setting is reapplied on every prebuild/build (and committed once in `package.json` / lockfile).
3. **Result:** Image-heavy sections (Home rails, Explore "Explore by event") load their photos on the built release APK. Combined with fix #1, recommendation photos require both the absolute URL and cleartext permission to appear.

## Changed files (7)

| File | +/− | Fix |
|---|---|---|
| `app/backend/apps/search/views.py` | +2 / −2 | #1 |
| `app/backend/apps/stories/serializers.py` | +2 / −1 | #2 |
| `app/mobile/src/services/storyService.ts` | +8 / −2 | #3 |
| `app/mobile/src/screens/UserProfileScreen.tsx` | −2 | #4 |
| `app/mobile/app.json` | +15 / −2 | #5 (maps) + #6 (cleartext) |
| `app/mobile/package.json` | +1 | #6 |
| `app/mobile/package-lock.json` | +52 (auto) | #6 |

## Note on the lockfile

The `package-lock.json` change (+52 lines) is auto-generated, not hand-written. Installing the `expo-build-properties` plugin (fix #6) pulled in that package plus its transitive dependencies; the lockfile records a version, resolved URL and integrity hash for each. It is a normal, expected side effect of adding the dependency.

## Deploy order

Backend merge + redeploy first, then build the APK (fix #3 depends on backend fix #2 being deployed).


mustafaocakxyz (approved):
LGTM

